PR #33, opened 2019-07-05
initialize kton tests

hammeWang:


srml/token/kton/src/lib.rs
```diff
@@ -115,7 +118,7 @@ decl_event!(
 
 
 decl_storage! {
-	trait Store for Module<T: Trait> as KtonBalances {
```
HackFisher: Are there storage compatible issues here? If true, does substrate have methods(e.g. migration design testing). To verify this question, maybe need do some testing

Issue #43, opened 2019-07-17
Proposals for kton module code refactoring


wuminzhe:
After understanding the kton module code, I think it involves too many responsibilities and can be separated more clearly: 

1. move lock rings logic to balances module, and then
2. increase kton balance in kton module through rings locked event.
3. move reward logic to a separate module maybe named "rewarding". (dividend-sharing logic here)
4. after kton balance changed, notify rewarding module to do reward.
5. SystemCurrency trait? I think this is not a general behavior.

wuminzhe:
kton's `deposit_creating` function should change name to `deposit_into_existing` according to `Currency` trait definition.

HackFisher:
Deprecated design, refer staking design in latest genepaper:

https://github.com/darwinia-network/rfcs/blob/master/paper/Darwinia_Genepaper_EN.pdf

PR #45, opened 2019-07-19
kton module code structure refactoring

wuminzhe:
#43 
compile success. 

extract the reward code to another two modules: **reward**, **balances-ext**.

**reward** contains the dividend-sharing logic.
**balances-ext** add deposit ring logic to balances module. I wanted to mix the new logic to the balances module, but failed. I think there is a way to extend substrate exsiting modules gracefully.

further adjustments and test needed.


wuminzhe:
the final refactoring result is:
1. extract reward logic to **reward** module.
2. reuse balances module to support kton currency.
2. create a delegate module to extend balances module.

wuminzhe:
I will create another one.

PR #49, opened 2019-07-26
 extract reward module

wuminzhe:
For more readable code, I extract the ring-lock and dividents-share logic to a standlone module.
And other develpers are also easier to understand.

### DIFF
--- a/srml/reward/Cargo.toml
+++ b/srml/reward/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
-name = "evo-staking"
+name = "evo-reward"
 version = "0.1.0"
-authors = ["hammeWang <dsw0602@foxmail.com>"]
+authors = ["Darwinia Network <hello@itering.com>"]
 edition = "2018"
 
 [dependencies]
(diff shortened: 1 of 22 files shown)


hammeWang:
1. Considered that reward mod is abstracted from KTON, trait `Currency` and `SystemRefund` can be removed from KTON mod.
2. Suggest that rename trait  `OnDilution` in `dsupport` mod, since we may reuse `srml::support::OnDilution` in the future when `Treasury` mod is finished.
3. please check all related tests in reward.

srml/reward/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+);
+
+decl_storage! {
+    trait Store for Module<T: Trait> as Reward {
+        pub DepositLedger get(deposit_ledger): map T::AccountId => Option<Deposit<RingBalanceOf<T>, KtonBalanceOf<T>, T::Moment>>;
+
+        pub SysAcc get(sys_acc) config(): T::AccountId;
+
+        // reward you can get per kton
+        pub RewardPerShare get(reward_per_share): RingBalanceOf<T>;
```
HackFisher: This should stay in KTON module.

srml/reward/src/lib.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        // reward already paid to each ktoner
+        pub RewardPaidOut get(reward_paid_out): map T::AccountId => i128;
```
HackFisher: This should stay in KTON module.

srml/reward/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "evo-reward"
```
HackFisher: reward should be renamed to Gringotts

HackFisher:
The dependency map should look like this.

RING:
- Currency

KTON:
- Currency
- Reward payout settlement related function(P3D alike)

Gringotts: Deposit management
- RING
- KTON

Staking:
- RING
- KTON

HackFisher:
Reference: https://github.com/evolutionlandorg/bank

wuminzhe:
Current hard dependency is:

    KTON: 
    REWARD: KTON and RING
    STAKING: KTON and RING

When kton balance changed, kton module notify reward module to update reward per share with OnAccountBalanceChanged.
If staking module issue new reward, it notify reward module to recode in pot. Reward module use to reward kton holder.

I think SysAcc, RewardPerShare and RewardPaidOut shoud be placed in a module independently because they not only denpend on kton, but also heavy depend on ring like staking. 

For me I like to extract two modules, one for P3D alike reward, one for deposit management.  The code will be so clear.

HackFisher:
The new dependency map should look like this.

RING ( impl Currency)：

KTON ( impl Currency):
OnBalanceChanged Trait

Reward (impl OnBalanceChanged): Reward payout settlement related function(P3D alike)
- RING
- KTON

Gringotts: Deposit management
- RING
- KTON

Staking: (TODO: 需要用新算法替换目前的for循环清算，目前算法是跟波卡一样)
- RING
- KTON

wuminzhe:
@HackFisher @hammeWang 
already split to reward and gringotts

PR #92, opened 2019-11-14
withdraw Ring/Kton automatically

AurevoirXavier:
update: change `LockableCurrency::Moment` from `BlockNumber` to `TimeStamp(u64)` for Ring/Kton


srml/balances/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+			Zero::zero()
+		}
+	}
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
+pub struct BalanceLock<Balance> {
+	pub id: LockIdentifier,
+	pub amount: Balance,
+	pub until: TimeStamp,
```
HackFisher: Can this be a generic Type, support both TimeStamp and BlockNumber?

AurevoirXavier: Changed.

srml/balances/src/lib.rs
```diff
@@ -0,0 +1,1208 @@
...
+		}
+		let locks = Self::locks(who);
+		if locks.is_empty() {
+			return Ok(());
+		}
+
+		let now = <timestamp::Module<T>>::now();
+		if locks
+			.into_iter()
+			.all(|l| now >= l.until || new_balance >= l.amount || !l.reasons.intersects(reasons))
```
HackFisher: This condition check here deserve careful checking when multiple locks exists at the same time, e.g. Stash balances and unbonding balances.

It seems that ``` new_balance >= l.amount ``` only checks the largest locked one, so additional locking will not take effect, maybe.

srml/balances/src/lib.rs
```diff
@@ -0,0 +1,1208 @@
...
+	// lock IDs, which means the number of runtime modules that intend to use and create locks.
+	// # </weight>
+	fn ensure_can_withdraw(
+		who: &T::AccountId,
+		_amount: T::Balance,
+		reasons: WithdrawReasons,
+		new_balance: T::Balance,
+	) -> Result {
+		if reasons.intersects(WithdrawReason::Reserve | WithdrawReason::Transfer)
+			&& Self::vesting_balance(who) > new_balance
```
HackFisher: Vesting Balance is still using BlockNumber, may break consistency

srml/balances/src/lib.rs
```diff
@@ -0,0 +1,1208 @@
...
+
+	// `amount` > `free_balance` is allowed
+	fn set_lock(
+		id: LockIdentifier,
+		who: &T::AccountId,
+		amount: T::Balance,
+		until: Self::Moment,
+		reasons: WithdrawReasons,
+	) {
+		let now = <timestamp::Module<T>>::now();
```
HackFisher: I found a problem that if the until is BlockNumber, it should not compare with a timestamp, right?

AurevoirXavier: In this case they’re `Moment`.

AurevoirXavier:
@HackFisher New design for `LockableCurrency`.

srml/kton/src/lib.rs
```diff
@@ -380,85 +382,46 @@ impl<T: Trait> Currency<T::AccountId> for Module<T> {
...
-				} else {
-					None
-				}
-			})
-			.collect::<Vec<_>>();
-		if let Some(lock) = new_lock {
-			locks.push(lock);
-		}
-		<Locks<T>>::insert(who, locks);
+	fn set_lock(who: &T::AccountId, amount: Self::Balance, id: Self::Id) {
```
HackFisher: ```Self::Id``` should be able to represent <id, util, reasons>.

Suggest to keep id, reasons in the trait,  and make ```Lock``` generalized to some enum which could represent complex locks such as ```  <util: ...> ```

srml/kton/src/lib.rs
```diff
@@ -380,85 +377,29 @@ impl<T: Trait> Currency<T::AccountId> for Module<T> {
...
-	) {
-		let now = <system::Module<T>>::block_number();
-		let mut new_lock = Some(BalanceLock {
-			id,
-			amount,
-			until,
-			reasons,
+		<Locks<T>>::mutate(who, |locks| {
+			locks.retain(|lock| lock.id != id || lock.valid(&now));
+			locks.push(BalanceLock { id, amount, reasons });
```
HackFisher: If there is a lock with same ID exists and also vaid, will it be retained with the new at the same time?

srml/support/src/types.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl<Moment> PartialEq for Id<Moment>
+where
+	Moment: PartialEq,
+{
+	fn eq(&self, other: &Self) -> bool {
+		match self {
+			Id::Staking(_) => match other {
+				Id::Staking(_) => true,
```
HackFisher: Are equal even the moment are different?

srml/staking/src/lib.rs
```diff
@@ -97,12 +97,12 @@ impl Default for ValidatorPrefs {
 }
 
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug)]
-pub enum StakingBalance<RingBalance, KtonBalance> {
-	Ring(RingBalance),
-	Kton(KtonBalance),
+pub enum StakingBalance<Ring, Kton> {
```
HackFisher: RingBalance looks good to me, since it represent a balance value.

srml/staking/src/lib.rs
```diff
@@ -440,52 +424,53 @@ decl_module! {
...
 
-					if !available_unbund_ring.is_zero() {
-						*active_ring -= available_unbund_ring;
-						unlocking.push(UnlockChunk {
-							value: StakingBalance::Ring(available_unbund_ring),
-							era,
-						});
+					if !available_unbond_ring.is_zero() {
+						*active_ring -= available_unbond_ring;
+//						TODO
```
HackFisher: TODO.

srml/staking/src/lib.rs
```diff
@@ -440,52 +424,53 @@ decl_module! {
...
+						<KtonPool<T>>::mutate(|k| *k -= unbond_kton);
 
-						*active_kton -= unbound_kton;
-
-						unlocking.push(UnlockChunk {
-							value: StakingBalance::Kton(unbound_kton),
-							era,
-						});
+						*active_kton -= unbond_kton;
+//						TODO
```
HackFisher: TODO

srml/staking/src/lib.rs
```diff
@@ -811,32 +785,26 @@ impl<T: Trait> Module<T> {
...
-
-			StakingBalance::Kton(_k) => T::Kton::set_lock(
-				STAKING_ID,
-				&ledger.stash,
-				ledger.total_kton,
-				T::BlockNumber::max_value(),
-				WithdrawReasons::all(),
-			),
-		}
+		//		TODO
```
HackFisher: TODO

srml/staking/src/lib.rs
AurevoirXavier: `Ring` and `Kton` are `StakingBalance`, I think it's clear enough.

srml/kton/src/lib.rs
AurevoirXavier: I think this's very similar to the old logic, but with different ordering in `Vec`.

if exists:
- old logic: update the exists lock
- update: remove the exists lock, push the new lock

srml/support/src/types.rs
AurevoirXavier: From the old logic of staking: It doesn't care about the `Moment` and it will update the lock anyway.

srml/kton/src/lib.rs
```diff
@@ -236,11 +232,42 @@ impl<T: Trait> Currency<T::AccountId> for Module<T> {
...
+					// can_withdraw = !lock.reasons.intersects(reasons);
+				}
+				Id::Unbonding(_) => {
+					// due to rule 3
+					can_withdraw_unbonding = !lock.valid_at(&now);
+				}
+			}
+		}
+
+		if can_withdraw_unbonding || can_withdraw_staking {
```
HackFisher: It is more clean and decent if this complex logic can be put in one composition Lock, such as Stakking_Unbund_Lock. Then, the balances module does not need to dependent on ID types. and do match checks here.

srml/balances/src/lib.rs
```diff
@@ -0,0 +1,1083 @@
...
+		amount: T::Balance,
+		reasons: WithdrawReasons,
+		new_balance: T::Balance,
+	) -> Result {
+		if reasons.intersects(WithdrawReason::Reserve | WithdrawReason::Transfer)
+			&& Self::vesting_balance(who) > new_balance
+		{
+			return Err("vesting balance too high to send value");
+		}
+		let (staking_lock, unbonding_locks) = Self::locks(who);
```
HackFisher: hmm... why not put staking lock and unbonding_lock in one lock, it is not a good idea that the balances module dependent on staking.

AurevoirXavier: All **Unbonding Locks** share one **Staking Lock**. If you define a lock which have a `staking_amount` filed and it also depends on staking logic. In substrate the `BalanceLock` was defined in balances module. There're two kinds of Currency on our chain, so I move the `BalanceLock to support module.

srml/balances/src/lib.rs
```diff
@@ -0,0 +1,1083 @@
...
+	fn set_lock(
+		who: &T::AccountId,
+		amount: Self::Balance,
+		at: Self::Moment,
+		reasons: Self::WithdrawReasons,
+	) -> Self::Balance {
+		let now = <timestamp::Module<T>>::now();
+		let mut expired_locks_amount = Self::Balance::default();
+
+		if at.saturated_into::<TimeStamp>() == 0 {
```
HackFisher: 我觉得Lockable Currency的lock机制不只是Staking模块会用到，其他模块也会用到。因此Balance模块中的Lock功能应该通用，特别的业务逻辑应该放到对应的业务模块中去：这里的staking_lock，unbounding_lock及其判断逻辑应该放在staking模块或support模块。

srml/balances/src/lib.rs
HackFisher: 我觉得Lockable Currency的lock机制不只是Staking模块会用到，其他模块也会用到。因此Balance模块中的Lock功能应该通用，特别的业务逻辑应该放到对应的业务模块中去：这里的staking_lock，unbounding_lock及其判断逻辑应该放在staking模块或support模块。

所以具体的结构还好，主要是这些特殊的lock具体实现机制不要放在Balance模块中

AurevoirXavier: There're some problems with your `check_valid`.

1. `new_balance >= staking_amount + (all)unbonding_amount`
2. `new_balance >= staking_amount` only need to check once

How about the `CompositeLock`?

```rust
CompositLock {
	staking_amount: Balance,
	locks: Vec<BalanceLock>,
}
```

HackFisher: Yes, this is better. (I forgot that there are multiple unbounding locks in staking).

And I suggest move all the checks to Lock trait(enum), CompositLock will require to implement 
```
check_valid(new_blance, time) && reasons.check ...
```
Then simply call lock.check_valid(new_blance, time) regardless what kind of it is in ```ensure withdraw```

AurevoirXavier: Yep, I wanted to move the whole `ensure_can_withdraw` logic into `LockableCurrency` at the beginning. It would be more concise.

And there's a [weird problem](https://github.com/darwinia-network/darwinia/pull/94#issuecomment-555331777) when I try to merge from develop branch. I can't compile the staking module but I can compile the darwinia.

Request review:
- [LockableCurrency](https://github.com/AurevoirXavier/darwinia/blob/9554a520375f7e0fd1546780e3322739b86a6734/srml/support/src/traits.rs#L25)
- [staking](https://github.com/AurevoirXavier/darwinia/blob/chainrelay/srml/staking/src/lib.rs)
- [balances](https://github.com/AurevoirXavier/darwinia/blob/chainrelay/srml/balances/src/lib.rs)
- [kton](https://github.com/AurevoirXavier/darwinia/blob/chainrelay/srml/kton/src/lib.rs)

srml/kton/src/lib.rs
```diff
@@ -374,71 +374,26 @@ impl<T: Trait> LockableCurrency<T::AccountId> for Module<T>
...
-					true
-				} else {
-					expired_locks_amount += lock.amount;
-					false
-				}
-			});
-		}
-
-		<Locks<T>>::insert(who, composite_lock);
+	fn update_lock(who: &T::AccountId, lock: Option<Self::Lock>) -> Self::Balance {
```
HackFisher:
```
	fn set_lock(
		id: LockIdentifier,
		who: &T::AccountId,
		new_lock: Option<Self::Lock>,
		reasons: WithdrawReasons,
	) {
		let mut locks = Self::locks(who).into_iter().filter_map(|l|
			if l.id == id {
				new_lock.take()
			} else {
				None
			}).collect::<Vec<_>>();
		if let Some(lock) = new_lock {
			locks.push(lock)
		}
		<Locks<T, I>>::insert(who, locks);
	}
// remove extend_lock
//	fn extend_lock

	fn remove_lock(
		id: LockIdentifier,
		who: &T::AccountId,
	) {
		let now = <system::Module<T>>::block_number();
		let locks = Self::locks(who).into_iter().filter_map(|l|
			if l.id != id {
				Some(l)
			} else {
				None
			}).collect::<Vec<_>>();
		<Locks<T, I>>::insert(who, locks);
	}
```

srml/staking/src/lib.rs
```diff
@@ -60,9 +61,9 @@ mod tests;
 const DEFAULT_MINIMUM_VALIDATOR_COUNT: u32 = 4;
 const MAX_NOMINATIONS: usize = 16;
 const MAX_UNSTAKE_THRESHOLD: u32 = 10;
-const MAX_UNLOCKING_CHUNKS: usize = 32;
+const MAX_UNLOCKING_CHUNKS: u32 = 32;
 const MONTH_IN_SECONDS: u32 = 2_592_000;
-const STAKING_ID: LockIdentifier = *b"staking ";
+const ERA_IN_SECONDS: TimeStamp = 300;
```
HackFisher: ```ERA_IN_SECONDS ``` seems been removed already.

srml/staking/src/lib.rs
```diff
@@ -1205,7 +1142,7 @@ impl<T: Trait> Module<T> {
 		CurrentEraStartSessionIndex::mutate(|v| {
 			*v = start_session_index;
 		});
-		let bonding_duration = T::BondingDuration::get();
+		let bonding_duration = (T::BondingDuration::get() / ERA_IN_SECONDS) as _;
```
HackFisher: ```ERA_IN_SECONDS ``` has already been removed? Check again.

srml/staking/src/lib.rs
```diff
@@ -313,8 +311,8 @@ pub trait Trait: timestamp::Trait + session::Trait {
 	/// Number of sessions per era.
 	type SessionsPerEra: Get<SessionIndex>;
 
-	/// Number of eras that staked funds must remain bonded for.
-	type BondingDuration: Get<EraIndex>;
+	/// Number of seconds that staked funds must remain bonded for.
+	type BondingDuration: Get<TimeStamp>;
```
HackFisher: Related to ```EraIndex```, check again.

srml/staking/src/lib.rs
```diff
@@ -469,7 +464,7 @@ decl_module! {
 		const SessionsPerEra: SessionIndex = T::SessionsPerEra::get();
 
 		/// Number of eras that staked funds must remain bonded for.
-		const BondingDuration: EraIndex = T::BondingDuration::get();
+		const BondingDuration: TimeStamp = T::BondingDuration::get();
```
HackFisher: Related to EraIndex, check again.

srml/support/src/types.rs
```diff
@@ -0,0 +1,206 @@
...
+use rstd::vec::Vec;
+use sr_primitives::{traits::SimpleArithmetic, RuntimeDebug};
+use srml_support::traits::WithdrawReasons;
+
+use super::traits::Locks as LocksTrait;
+
+pub type TimeStamp = u64;
+
+#[derive(Clone, Default, PartialEq, Encode, Decode, RuntimeDebug)]
+pub struct Locks<Balance, Moment>(pub Vec<CompositeLock<Balance, Moment>>);
```
HackFisher: Questions here

srml/staking/src/lib.rs
```diff
@@ -519,12 +514,15 @@ decl_module! {
...
+			ensure!(promise_month <= 36, "months at most is 36.");
 			let controller = Self::bonded(&stash).ok_or("not a stash")?;
-			let ledger = Self::ledger(&controller).ok_or("not a controller")?;
+			let mut ledger = Self::ledger(&controller).ok_or("not a controller")?;
 			match value {
 				 StakingBalance::Ring(r) => {
 					let stash_balance = T::Ring::free_balance(&stash);
+					let expired_locks_ring = T::Ring::update_lock(&stash, None);
+					// TODO: check underflow?
+					ledger.total_ring -= expired_locks_ring;
```
HackFisher: Why update expired locks here and remove from total_ring. I remember that unbounding balances is not included in total ring anymore.

srml/staking/src/lib.rs
```diff
@@ -533,6 +531,8 @@ decl_module! {
 				},
 				StakingBalance::Kton(k) => {
 					let stash_balance = T::Kton::free_balance(&stash);
+					let expired_locks_kton = T::Kton::update_lock(&stash, None);
+					ledger.total_kton -= expired_locks_kton;
```
HackFisher: 同上.

srml/staking/src/lib.rs
```diff
@@ -552,54 +552,66 @@ decl_module! {
...
 				active_deposit_ring,
+				total_kton,
 				active_kton,
-				unlocking,
 				..
 			} = &mut ledger;
 
 			ensure!(
-				unlocking.len() < MAX_UNLOCKING_CHUNKS,
+				T::Ring::locks_count(stash) + T::Kton::locks_count(stash) < MAX_UNLOCKING_CHUNKS,
```
HackFisher: TODO: Refactor

srml/staking/src/lib.rs
```diff
@@ -552,54 +552,66 @@ decl_module! {
...
 				"can not schedule more unlock chunks"
 			);
 
-			let era = Self::current_era() + T::BondingDuration::get();
-
+			let at = <timestamp::Module<T>>::now().saturated_into::<TimeStamp>() + T::BondingDuration::get();
```
HackFisher: Check: why change from ```Self::current_era() ``` to ```<timestamp::Module<T>>::now()```

srml/staking/src/lib.rs
```diff
@@ -552,54 +552,66 @@ decl_module! {
...
+						let expired_locks_ring = T::Ring::update_lock(
+							stash,
+							Some(CompositeLock::Unbonding(Lock {
+								amount: available_unbond_ring,
+								at,
+								reasons: WithdrawReasons::all()
+							})),
+						);
+						// TODO: check underflow?
+						*total_ring -= expired_locks_ring;
```
HackFisher: Check: unbounding is not included in total_ring.

srml/staking/src/lib.rs
```diff
@@ -552,54 +552,66 @@ decl_module! {
...
+						let expired_locks_kton = T::Kton::update_lock(
+							stash,
+							Some(CompositeLock::Unbonding(Lock {
+								amount: unbond_kton,
+								at,
+								reasons: WithdrawReasons::all(),
+							})),
+						);
+						// TODO: check underflow?
+						*total_kton -= expired_locks_kton;
```
HackFisher: Check: unbounding is not included in total_ring.

srml/staking/src/mock.rs
```diff
@@ -156,11 +171,12 @@ impl kton::Trait for Test {
 	type Event = ();
 	type OnMinted = ();
 	type OnRemoval = ();
+	type Locks = Locks<Balance, TimeStamp>;
 }
 
 parameter_types! {
 	pub const SessionsPerEra: SessionIndex = 3;
-	pub const BondingDuration: EraIndex = 3;
+	pub const BondingDuration: TimeStamp = 60;
```
HackFisher: Check EraIndex

srml/staking/src/phragmen.rs
```diff
@@ -0,0 +1,505 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Rust implementation of the Phragmén election algorithm. This is used in several SRML modules to
+//! optimally distribute the weight of a set of voters among an elected set of candidates. In the
+//! context of staking this is mapped to validators and nominators.
+//!
+//! The algorithm has two phases:
+//!   - Sequential phragmen: performed in [`elect`] function which is first pass of the distribution
+//!     The results are not optimal but the execution time is less.
```
HackFisher: Check: this file has been removed!

node/cli/src/lib.rs
```diff
@@ -219,23 +213,21 @@ where
 			};
 
 			match ChainSpec::from(config.chain_spec.id()) {
-				Some(ref c) if c == &ChainSpec::Development || c == &ChainSpec::LocalTestnet => {},
```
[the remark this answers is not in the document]
AurevoirXavier: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=77ffce8526a5d30146e36ae0ba758ced

srml/staking/src/lib.rs
AurevoirXavier: The lock require a `TimeStamp` to chcek if expired.

srml/staking/src/lib.rs
AurevoirXavier: These was the expired locks. In a 'Finish Unbond' state.

srml/staking/src/lib.rs
AurevoirXavier: These are the expired locks. In a 'Finish Unbond' state.

srml/staking/src/lib.rs
```diff
@@ -1142,7 +1141,10 @@ impl<T: Trait> Module<T> {
 		CurrentEraStartSessionIndex::mutate(|v| {
 			*v = start_session_index;
 		});
-		let bonding_duration = (T::BondingDuration::get() / ERA_IN_SECONDS) as _;
+		let bonding_duration = {
```
HackFisher: Better to rename to ```bonding_era``` to avoid term misunderstanding

PR #97, opened 2019-11-20
Refactoring LockableCurrency

HackFisher:
- Keep the logic same with original ```LockableCurrency``` except the withdraw lock abstraction changes.
- Keep Lock Identifier, one lock for one ID.
- Refactoring, the latest changes are kind to over complex.

TODO:
- remove ```total_kton``` and ```total_ring```


HackFisher:
https://oribenshir.github.io/afternoon_rusting/blog/enum-and-pattern-matching-part-1

https://oribenshir.github.io/afternoon_rusting/blog/enum-and-pattern-matching-part-2

srml/support/src/lib.rs
```diff
@@ -1,5 +1,176 @@
+#![recursion_limit = "128"]
 #![cfg_attr(not(feature = "std"), no_std)]
 #![feature(drain_filter)]
 
-pub mod traits;
-pub mod types;
```
AurevoirXavier: Recommend to seperate into smaller modules. The support things are unrelated. We provide these to other people to use.

If you want to re-export add a `pub use` or put them in a `prelude` module.

srml/staking/src/lib.rs
```diff
@@ -599,16 +587,9 @@ decl_module! {
...
-							Some(CompositeLock::Unbonding(Lock {
-								amount: unbond_kton,
-								at,
-								reasons: WithdrawReasons::all(),
-							})),
-						);
+						(*kton_detail_lock).unlocking.push(UnlockChunk { value: unbond_kton, until: at });
 						// TODO: check underflow?
-						*total_kton -= expired_locks_kton;
+						*total_kton -= unbond_kton;
```
AurevoirXavier: `total_kton - active_kton = unbonding_kton`

Does it means no Kton in unbonding? 

And `total_kton` always equals to `active_kton` now:
	- both of them are increased in `bond_helper_in_kton` at the same time
	- both of them are decreased in `unbond` at the same time

srml/staking/src/lib.rs
```diff
@@ -64,6 +64,8 @@ const MAX_UNSTAKE_THRESHOLD: u32 = 10;
 const MAX_UNLOCKING_CHUNKS: u32 = 32;
 const MONTH_IN_SECONDS: u32 = 2_592_000;
 
+const STAKING_ID: LockIdentifier = *b"staking ";
```
AurevoirXavier: `LockIdentifier` is for differentiate the locks. If we custom a `enum_id/lock` we could build a more powerful lock.

srml/support/src/traits.rs
```diff
@@ -1,78 +0,0 @@
...
-pub trait OnMinted<Balance> {
-	fn on_minted(value: Balance);
-}
-
-pub trait OnAccountBalanceChanged<AccountId, Balance> {
-	fn on_changed(who: &AccountId, old: Balance, new: Balance);
-}
-
-/// A more powerful lockable currency.
-pub trait LockableCurrency<AccountId>: Currency<AccountId> {
```
AurevoirXavier: In support module, I want to make things more generic so other people can reuse this instead design a new trait like us.

srml/support/src/lib.rs
```diff
@@ -1,5 +1,176 @@
...
+#[derive(Clone, PartialEq, Eq, Encode, Decode, RuntimeDebug)]
+pub struct StakingAndUnbondingLock<Balance, Moment> {
+	pub staking_amount: Balance,
+	// now >= l.until || new_balance >= l.amount
+	/// Any balance that is becoming free, which may eventually be transferred out
+	/// of the stash (assuming it doesn't get slashed first).
+	pub unlocking: Vec<UnlockChunk<Balance, Moment>>,
+}
+
+impl<Balance: Default, Moment: Default> Default for StakingAndUnbondingLock<Balance, Moment> {
```
AurevoirXavier: Simply `#[derive(Default)]`.

srml/support/src/lib.rs
```diff
@@ -1,5 +1,176 @@
...
+//}
+
+impl<Balance, Moment> StakingAndUnbondingLock<Balance, Moment>
+where
+	Balance: Clone + Copy + Default + SimpleArithmetic,
+	Moment: Clone + Copy + PartialOrd,
+{
+	fn can_withdraw(&self, at: Moment, new_balance: Balance) -> bool {
+		// TODO: Is it correct to use clone here?
+		let mut locked_amount = self.staking_amount.clone();
```
AurevoirXavier: `staking_amount` has a `Copy` bound. If we want to take its ownership, it'll be cloned automactically.

srml/staking/src/lib.rs
HackFisher: enum is used for composite of different lock structs, not for different locks.
We could have different ```LockIdentifer``` for same lock struct.

```LockIdentifer``` is a more straight and simple way to orgnize locks for each account.

srml/support/src/traits.rs
HackFisher: Moving the traits to mod trait to follow the substrate conventions.

I put them on the same file because current support module are quite simple, which can be splitted to other file if code go large.

srml/staking/src/lib.rs
HackFisher: I guess total_kton can be removed now. It is now related to slash now, but I think slash should use active_kton/active_ring instead.

PR #100, opened 2019-11-21
Patch: for #97

AurevoirXavier:
Patch: for #97


HackFisher:
I'm still working on #97 , stay tuned.

Issue #60, opened 2019-08-09
record: ugly hacking in staking


hammeWang:
#### Problem Discription
use 10**9 instead of some constant like `COIN` to represent the unit of RING. Really ugly hacking.

constant `COIN` is defined in `node_runtime`, it can not be used in staking because of cycling dependency.

PR #102, opened 2019-11-22
update: sync tests and runtime primitives

AurevoirXavier:
- Bugs Fixes: https://github.com/darwinia-network/darwinia/pull/102/commits/920b3380b2c6bb621d0e0eae6d110469ea0fe748 
- Logic Changes: https://github.com/darwinia-network/darwinia/pull/102/commits/b0cdc1e6f89f945bfbee9f9ab755e106cd02ff4b, https://github.com/darwinia-network/darwinia/pull/102/commits/129416a2d4ec67e16a3e673f8dff297a861cfda4
- Types Changes: https://github.com/darwinia-network/darwinia/pull/102/commits/9f80a33ca9128cd674d4a24c3c37fb64aeee7365
	- Fix: #60
- Tests Updates: ...


node/runtime/src/constants.rs
```diff
@@ -20,7 +20,10 @@
 pub mod currency {
 	use node_primitives::Balance;
 
-	pub const MILLICENTS: Balance = 1_000_000_000;
+	pub const NANO: Balance = 1;
+	pub const MICRO: Balance = 1_000 * NANO;
+	pub const MILLI: Balance = 1_000 * MICRO;
+	pub const MILLICENTS: Balance = 1_000 * MILLI;
```
HackFisher: Clarification:

1 Nano = 10^-9 COIN(RING/KTON), 1 Dollar = 1 COIN(RING/KTON).

So we may not user terms ```MILLICENTS, CENTS, DOLLARS ``` anymore.

srml/kton/src/mock.rs
```diff
@@ -1,27 +1,33 @@
+pub use node_runtime::constants::currency::MILLICENTS;
```
HackFisher: It is not good idea for srml to dependent on node_runtime.

Because if some other projects (or runtime) want to use this srml, they will have to dependent on darwinia runtime.

AurevoirXavier: No, this only define in `[dev-dependencies]`. Just for testing env.

node/runtime/src/constants.rs
AurevoirXavier: Ok, I'll update the name.

AurevoirXavier:
slash tests (hanging): https://github.com/darwinia-network/darwinia/commit/12d92d0563e74afeabe0611a937965611e8d4f87#commitcomment-36138167

Issue #117, opened 2019-11-26
sr-eth-primitives no_std support error?


HackFisher:
```
error: failed to run custom build command for `node-runtime v0.1.0 (/Users/denny/github.com/darwinia-network/darwinia/node/runtime)`

Caused by:
  process didn't exit successfully: `/Users/denny/github.com/darwinia-network/darwinia/target/release/build/node-runtime-f7eaf2f8b8370061/build-script-build` (exit code: 1)
--- stdout
Executing build command: "cargo" "build" "--target=wasm32-unknown-unknown" "--manifest-path=/Users/denny/github.com/darwinia-network/darwinia/target/release/wbuild/node-runtime/Cargo.toml" "--release"

--- stderr
   Compiling wasm-build-runner-impl v1.0.0 (/Users/denny/github.com/darwinia-network/darwinia/target/release/wbuild-runner/node-runtime)
    Finished release [optimized] target(s) in 0.29s
     Running `/Users/denny/github.com/darwinia-network/darwinia/target/release/wbuild-runner/node-runtime/target/release/wasm-build-runner-impl`
   Compiling keccak-hash v0.4.1
   Compiling sr-io v2.0.0 (https://github.com/darwinia-network/substrate.git?branch=darwinia-develop#d2c4b0db)
error[E0152]: duplicate lang item found: `panic_impl`.
   --> /Users/denny/.cargo/git/checkouts/substrate-a7fa553ead635512/d2c4b0d/core/sr-io/src/lib.rs:737:1
    |
737 | / pub fn panic(info: &core::panic::PanicInfo) -> ! {
738 | |     unsafe {
739 | |         let message = rstd::alloc::format!("{}", info);
740 | |         misc::print_utf8(message.as_bytes());
741 | |         core::intrinsics::abort()
742 | |     }
743 | | }
    | |_^
    |
    = note: first defined in crate `std` (which `parity_scale_codec` depends on).

error[E0152]: duplicate lang item found: `oom`.
   --> /Users/denny/.cargo/git/checkouts/substrate-a7fa553ead635512/d2c4b0d/core/sr-io/src/lib.rs:747:1
    |
747 | / pub extern fn oom(_: core::alloc::Layout) -> ! {
748 | |     static OOM_MSG: &str = "Runtime memory exhausted. Aborting";
749 | |
750 | |     unsafe {
...   |
753 | |     }
754 | | }
    | |_^
    |
[...]

PR #119, opened 2019-11-27
fix error while compling no_std wasm

hammeWang:


hammeWang:
close https://github.com/darwinia-network/darwinia/issues/117

hammeWang:
@AurevoirXavier what's your point? sorry I don't get you.

AurevoirXavier:
@hammeWang **default** contains std, so we disable it. But we add **serialize** which contains std. `default-features = false` will be overwrited.

hammeWang:
@AurevoirXavier yep, that's the reason `feature serailze` need to be removed.

PR #141, opened 2019-12-05
fix balances tests

readlnh:

PR #155, opened 2019-12-10
Develop

AurevoirXavier:
- fix: 
	1. #136
	2. all timestamp with milliseconds format
	3. use `u64` for all `Moment`

- update: 
	1. staking module
	2. `StakingLock` using generic `Moment`

### DIFF
--- a/srml/balances/src/tests.rs
+++ b/srml/balances/src/tests.rs
@@ -23,10 +23,10 @@ use support::{
 use system::RawOrigin;
 use transaction_payment::ChargeTransactionPayment;
 
-use darwinia_support::{LockIdentifier, WithdrawReason, WithdrawReasons, WithdrawLock, NormalLock, LockableCurrency};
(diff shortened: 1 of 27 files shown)


srml/balances/src/tests.rs
```diff
@@ -84,13 +83,13 @@ fn lock_removal_should_work() {
...
+				ID_1,
+				&1,
 				WithdrawLock::Normal(NormalLock {
-					amount: u64::max_value(), 
+					amount: u64::max_value(),
 					until: u64::max_value(),
-				}), 
-				WithdrawReasons::all()
+				}),
+				WithdrawReasons::all(),
```
HackFisher: Are these code format rules change?

AurevoirXavier: These file had been modified by @readlnh in #141.

I think we should make a git-hook to keep `cargo fmt`.

PR #159, opened 2019-12-11
Fix testing wrong data

HackFisher:
But mixhash verify on ropsten still fails if check, so still omit the check on ropsten.


srml/eth-relay/src/tests.rs
```diff
@@ -68,7 +68,7 @@ fn verify_receipt_proof() {
 				hash: Some(H256::from(hex!("f1a5bc27877e219b859b0bb1f2f440134553019f9bb5a2eca7a4703263e736c9"))),
 			};
 
-			EthRelay::init_genesis_header(&header, 0x624c22d93f8e59_u64);
+			EthRelay::init_genesis_header(&header, 0x624c22d93f8e59_u64).expect("Initialize Genesis Failed.");
```
AurevoirXavier: use `assert_ok!` and it will raise the error detail in console

srml/eth-relay/src/tests.rs
```diff
@@ -126,7 +126,7 @@ fn relay_header() {
 		};
 
 
-		EthRelay::init_genesis_header(&header1, 0x624c22d93f8e59_u64);
+		EthRelay::init_genesis_header(&header1, 0x624c22d93f8e59_u64).expect("Initialize Genesis Failed.");
```
AurevoirXavier: use `assert_ok!` and it will raise the error detail in console

srml/eth-relay/src/lib.rs
```diff
@@ -201,7 +210,11 @@ impl<T: Trait> Module<T> {
 	/// 2. proof of pow (mixhash)
 	/// 3. challenge
 	fn verify_header(header: &EthHeader) -> Result {
-		// TODO: check parent hash,
+		ensure!(
+			header.hash() == header.re_compute_hash(),
+			"Header Hash Check Failed - Hash NOT MATCH",
```
AurevoirXavier: I think `Header Hash - MISMATCHED` is better.

Could you please replace all `NOT MATCHED` with `MISMATCHED`?

srml/eth-relay/src/lib.rs
```diff
@@ -157,8 +158,14 @@ decl_event! {
 
 impl<T: Trait> Module<T> {
 	// TOOD: what is the total difficulty for genesis/begin header
-	pub fn init_genesis_header(header: &EthHeader, genesis_difficulty: u64) {
+	pub fn init_genesis_header(header: &EthHeader, genesis_difficulty: u64) -> result::Result<(), &'static str> {
 		let header_hash = header.hash();
+
+		ensure!(
+			header_hash == header.re_compute_hash(),
+			"Header Hash Check Failed - Hash NOT MATCH",
```
AurevoirXavier: I think Header Hash - MISMATCHED is better.

Could you please replace all NOT MATCHED with MISMATCHED?

PR #164, opened 2019-12-13
Add parse ring redeem receipt test

wuminzhe:


srml/eth-relay/src/tests.rs
```diff
@@ -1,16 +1,35 @@
...
-use mock::{EthRelay, ExtBuilder, System};
+use hex_literal::hex;
+use rustc_hex::FromHex;
+use sha3::Keccak256;
+use std::str::FromStr;
+use std::u64;
+use primitives::crypto::Ss58Codec;
+use node_primitives::AccountId;
+
+pub fn to_hex_string(bytes: Vec<u8>) -> String {
```
HackFisher: Why not directly convert using:

```
let raw_sub_key : Vec<u8> = result.params[3].value.clone().to_bytes().expect("Param Convert to Bytes Failed.");

			let darwinia_account: AccountId32 = AccountId32::try_from(raw_sub_key.as_slice()).expect("Account Parse Failed.");
```

srml/eth-relay/src/tests.rs
```diff
@@ -103,6 +122,8 @@ fn relay_header() {
 		// https://api-ropsten.etherscan.io/api?module=proxy&action=eth_getBlockByNumber&tag=0x69226b&boolean=true&apikey=YourApiKeyToken
 		// https://jsoneditoronline.org/
 
+		type DAG = LightDAG<EthereumPatch>;
```
AurevoirXavier: Already define in `crate` just `use crate::*`.

srml/eth-relay/Cargo.toml
```diff
@@ -20,6 +20,9 @@ rlp = { package = "rlp", git = "https://github.com/darwinia-network/parity-commo
 ethash = { path = "../../core/ethash", default-features = false }
 merkle-patricia-trie = { path = "../../core/merkle-patricia-trie", default-features = false }
 
+hex-literal = "0.2.1"
```
AurevoirXavier: Only use for testing?

Please move to `[dev-dependencies]`.

srml/eth-relay/Cargo.toml
```diff
@@ -20,6 +20,9 @@ rlp = { package = "rlp", git = "https://github.com/darwinia-network/parity-commo
...
+node-primitives = { path = "../../node/primitives" }
```
AurevoirXavier: Same as above.

srml/eth-relay/src/tests.rs
HackFisher: Please refer: https://github.com/darwinia-network/darwinia/blob/develop/srml/eth-backing/src/lib.rs#L313

This test can also been kept as alternative sample for experimental purpose. cause ```String``` seems can not be used in substrate module.

AurevoirXavier: In fact `String` is `Vec<u8>`.

And yes, substrate use `Vec<u8>` to represent the `String`.

But I found `to_owned` can convert the `&str` to `String` in substrate module. https://github.com/darwinia-network/darwinia/blob/develop/srml/eth-backing/src/lib.rs#L239

Furthermore, we can run the test in `std`, `String` can be brought into this mod.

```rust
#[cfg(all(feature = "std", test))]
mod tests;
```

PR #167, opened 2019-12-16
Develop

AurevoirXavier:


HackFisher:
> @HackFisher
> 
> All fees reset to 0 except `TransferFee`, please config them in `node/runtime/src/lib.rs`.

Prefer to setup a non zero fee for testing purpose

PR #170, opened 2019-12-17
Patch

AurevoirXavier:


srml/eth-backing/src/lib.rs
```diff
@@ -266,6 +196,79 @@ decl_module! {
 	}
 }
 
+impl<T: Trait> Module<T> {
+	fn parse_proof(proof: &EthReceiptProof, event_name: &str) -> result::Result<(T::AccountId, Balance), &'static str> {
```
HackFisher: RING and KTON happen to have the same event parameters, this is not general, suggest to rename it to ```parse_token_redeem_proof```

AurevoirXavier: Yes, I found `redeem_deposit`.

Issue #144, opened 2019-12-06
EthRelay: Add white list two factor protection option for production env


HackFisher:
For checking workers who can summit block header and receipt proof.

PR #185, opened 2019-12-24
Add white list two factor protection for eth relay module

wuminzhe:
#144

### DIFF
--- a/srml/eth-relay/src/lib.rs
+++ b/srml/eth-relay/src/lib.rs
@@ -158,9 +158,10 @@ decl_module! {
 		pub fn remove_authority(origin, who: T::AccountId) -> Result {
 			let _me = ensure_root(origin)?;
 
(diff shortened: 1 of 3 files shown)


srml/eth-relay/src/lib.rs
```diff
@@ -138,6 +144,35 @@ decl_module! {
...
+				<Authorities<T>>::mutate(|l| l.push(who));
+			}
+
+			Ok(())
+		}
+
+		pub fn remove_authority(origin, who: T::AccountId) -> Result {
+			let _me = ensure_root(origin)?;
+
+			if Self::authorities().contains(&who) {
```
AurevoirXavier:
```rust
if let Some(i) = Self::authorities().into_iter().position(|who_| who_ == who) {
	<Authorities<T>>::mutate(|s| s.remove(i));
}
```

PR #202, opened 2019-12-26
Develop

AurevoirXavier:
fix: im-online tests
update: more reasonable slash fraction

---

~~ethash tests will fail the CI~~ - #203


srml/im-online/src/mock.rs
```diff
@@ -127,7 +127,7 @@ parameter_types! {
 impl session::Trait for Runtime {
 	type ShouldEndSession = session::PeriodicSessions<Period, Offset>;
 	type OnSessionEnding = session::historical::NoteHistoricalRoot<Runtime, TestOnSessionEnding>;
-	type SessionHandler = (ImOnline);
+	type SessionHandler = (ImOnline,);
```
HackFisher: Is this tuple or something else?

AurevoirXavier: Yes, it need to be a tuple

PR #229, opened 2020-01-06
Add note to build without sccache

yanganto:
- add a note to avoid to use `sccache`


AurevoirXavier:
move to a *Precautions* part or *Compile QA*?

yanganto:
Thanks

PR #237, opened 2020-01-09
Trival/bootstrap

yanganto:


yanganto:
The result of bootstrap.sh on Ubuntu CI server.
https://travis-ci.org/darwinia-network/darwinia/jobs/634595251?utm_medium=notification&utm_source=github_status

AurevoirXavier:
Is this only for CI?
What if I use other package manage on those OS?

yanganto:
This is not for CI.  
It is just a script help developers put git hooks and set up the developing environment.
I do not run Ubuntu,  but CI does.
So the link shows the result I use CI machine to check this script.
After the script running on CI machine, I drop the commit which trigger CI machine to run the script.

Currently, the script only support Ubuntu, Arch & Mac (this is cover most developers in this project now), the other OS will be detected and show message for contribution welcome.

AurevoirXavier:
Okay, but I use **nixpkgs** on macOS and ArchLinux instead **homebrew**, **pacman**. Could you provide an option to skip the dependencies setup step?

yanganto:
Thanks for the review.  The `--fast` option is provided to skip the package installation.

PR #239, opened 2020-01-09
Fix documentation

yanganto:
- fix document generate script in README
- tab format in Cargo.toml at root
- fix cli doc string with correct crate name


node/cli/src/lib.rs
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
-//! Substrate CLI library.
+//! Darwminia CLI library.
```
AurevoirXavier: Darwinia

yanganto: Thanks.

PR #241, opened 2020-01-13
Preset the keys of Alice & Bob

### LINKED ISSUE #240: Replace the  Initial Accounts with Custom Accounts
replace the Alice, Bob in `initial_accounts` with custom accounts:
https://github.com/darwinia-network/darwinia/blob/develop/node/cli/src/chain_spec.rs#L316

also the eth-relay whitelist:
https://github.com/darwinia-network/darwinia/blob/develop/node/cli/src/chain_spec.rs#L203

yanganto:
- use predefined keys for Alice and Bob
- fix #240


yanganto:
Alice secret is generated here:
https://substrate.dev/rustdocs/v1.0/src/substrate_keyring/sr25519.rs.html#79

yanganto:
Should we change the account used in eth-relay?
Then the keys of Alice and Bob may keep in the original.

node/cli/src/chain_spec.rs
```diff
@@ -185,25 +244,23 @@ pub fn darwinia_genesis(
 ) -> GenesisConfig {
 	let endowed_accounts: Vec<AccountId> = endowed_accounts.unwrap_or_else(|| {
```
AurevoirXavier: If the `endowed_accounts` is `None`, this will give a testnet default accounts set.

Okay, move the origin code to `testnet_genesis`. And Keep the `darwinia_genesis`.

Only the `gen` function use `darwinia_genesis`. Other testnet chains run with `testnet_genesis`.

node/cli/src/chain_spec.rs
```diff
@@ -185,25 +244,23 @@ pub fn darwinia_genesis(
...
 			get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
 		]
 	});
 
-	let eth_relay_authorities: Vec<AccountId> = vec![
-		get_account_id_from_seed::<sr25519::Public>("Alice"),
-		get_account_id_from_seed::<sr25519::Public>("Bob"),
-	];
+	let eth_relay_authorities: Vec<AccountId> =
+		vec![initial_authorities[0].clone().1, initial_authorities[1].clone().1];
```
AurevoirXavier: Also here.

yanganto: A flag `is_test_net` is used to check the config is used under a test net or not.
In test net, the config will use original random generate keys for Alice and Bob.

yanganto:
Thanks for review.

PR #246, opened 2020-01-17
Try Time Wrap

### LINKED ISSUE #245: Unexpected Epoch Change When All Nodes Down
Due to #244, we've to stop all the validator(eleted) nodes and restart them in native mode. But `Unexpected Epoch Change` happended when all nodes down.

According to https://github.com/paritytech/substrate/issues/4464, **hard spoon** is the only option to recover the chain.

AurevoirXavier:
Try to fix #245


core/consensus/slots/src/slots.rs
```diff
@@ -0,0 +1,178 @@
...
+		}
+		.as_millis() as u64)
+			/ slot_duration
+	}
+}
+
+/// Returns the duration until the next slot, based on current duration since
+pub fn time_until_next(now: Duration, slot_duration: u64) -> Duration {
+	// ICEFROG HOTFIX: poll the slot 10 times as often since we might be in a time warp.
+	let slot_duration = slot_duration / 20;
```
andresilva:
```suggestion
	let slot_duration = slot_duration / 3;
```

AurevoirXavier:
https://github.com/paritytech/substrate/issues/4464#issuecomment-575573552

AurevoirXavier:
- Roll Back Point: 328600
- FROK TIMESTAMP: 1579057500
- REVIVE TIMESTAMP: 1579320000000 1/18/2020 12:00:00

Issue #222, opened 2020-01-04
Slash Strategy


readlnh:
Slashing is performed according the exposure (or more precisely, historical exposure), not the current balance in Substrate. However, our slashing is performed according to the current balance, not the exposure, is this what we want?

AurevoirXavier:
Yes, go #198 for more detail

AurevoirXavier:
Although we store **RING/KTON** in `Exposure`, **Power** still needed.

```rust
// This might reasonably saturate and we cannot do much about it. The sum of
// someone's stake might exceed the balance type if they have the maximum amount
// of balance and receive some support. This is super unlikely to happen, yet
// we simulate it in some tests.
```

AurevoirXavier:
for slash currency percisely, `Exposure` shoule be a snapshot of ledger at that time which build from `Support`

```rust
/// Means a particular `AccountId` was backed by `Votes` of a nominator's stake.
pub struct PhragmenStakedAssignment<AccountId, RingBalance, KtonBalance> {
	pub account_id: AccountId,
	pub ring_balance: RingBalance,
	pub kton_balance: KtonBalance,
}

/// A structure to demonstrate the phragmen result from the perspective of the candidate, i.e. how
/// much support each candidate is receiving.
///
/// This complements the [`PhragmenResult`] and is needed to run the equalize post-processing.
///
/// This, at the current version, resembles the `Exposure` defined in the staking SRML module, yet
/// they do not necessarily have to be the same.
#[derive(Default, RuntimeDebug)]
#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
pub struct Support<AccountId, RingBalance, KtonBalance> {
	/// The amount of support as the effect of self-vote.
	pub own_votes: Votes,
	pub own_ring: RingBalance,
	pub own_kton: KtonBalance,
	/// Total support.
	pub total: Votes,
	pub total_ring: RingBalance,
	pub total_kton: KtonBalance,
	/// Support from voters.
	pub others: Vec<PhragmenStakedAssignment<AccountId, RingBalance, KtonBalance>>,
}
```

PR #267, opened 2020-02-14
New Slash

AurevoirXavier:


AurevoirXavier:
Work on a problem about precision, may be something wrong with `Rational32`.

primitives/phragmen/src/lib.rs
```diff
@@ -85,7 +101,14 @@ pub struct Edge<AccountId> {
 pub type PhragmenAssignment<AccountId> = (AccountId, Perbill);
 
 /// Means a particular `AccountId` was backed by `Votes` of a nominator's stake.
-pub type PhragmenStakedAssignment<AccountId> = (AccountId, Votes);
+#[derive(RuntimeDebug)]
```
HackFisher: Votes, Power and Balances might have different meaning, need double checking here if we are goting to replace Votes.

primitives/phragmen/src/lib.rs
```diff
@@ -141,12 +168,13 @@ pub fn elect<AccountId, FS>(
 	initial_candidates: Vec<AccountId>,
 	initial_voters: Vec<(AccountId, Vec<AccountId>)>,
 	power_of: FS,
-	total_power: Votes,
 ) -> Option<PhragmenResult<AccountId>>
 where
 	AccountId: Default + Ord + Member,
-	for<'r> FS: Fn(&'r AccountId) -> Votes,
+	for<'r> FS: Fn(&'r AccountId) -> Power,
```
HackFisher: Is it good to replace Votes with Power?

frame/staking/src/lib.rs
```diff
@@ -539,33 +611,50 @@ pub struct Nominations<AccountId> {
...
+where
+	RingBalance: HasCompact,
+	KtonBalance: HasCompact,
+{
+	/// The validator's own stash that is exposed.
+	#[codec(compact)]
+	pub own_ring_balance: RingBalance,
+	#[codec(compact)]
+	pub own_kton_balance: KtonBalance,
+	pub own_power: Power,
```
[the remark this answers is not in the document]
AurevoirXavier: Yes.

Power is a dynamic value, we can't use it for slashing. So in previous version we use a perbill which calculated from `own_power/total_power` and apply it to active_ring/kton, but this will cause a problem which  was mentioned in #222 (if staker bond or unbond the active_ring/kton will changed immediately). We need a snapshot of ledger to perform the slashing at that moment. Also we should slash their unlocking items (if their active currency not enough for slashing).

primitives/phragmen/src/lib.rs
AurevoirXavier: Yes, we pass `power_of` to this function.

primitives/phragmen/src/lib.rs
AurevoirXavier: Yes. 

Total power is `1_000_000_000u32`.
The votes is `u64` which use for electing. There're some `mul` in electing which will overflow `Power: u32`. The `DEN` is `u64::max_value()`, if we use `1_000_000_000` there will be some accuracy problem (you can change it to `1_000_000_000` and run `cargo test` in phragmen folder). 
The balance is a cached data for slashing.

Issue #171, opened 2019-12-17
staking: automatically generate GenesisTime to the genesis block's time?


HackFisher:
There is a `GenesisTime` config in staking module, can it be removed?

readlnh:
You mean config in `mock.rs`? As we don't use `GenesisTime` in tests now,  there's no effects if we remove the config and change it to `type GenesisTime = ();`. However, i'm not sure wheather it is needed if we implement the [current_total_payout_for_duration](https://github.com/darwinia-network/darwinia/blob/develop/srml/staking/src/mock.rs#L484) later.

AurevoirXavier:
The config in runtime. We need the GenesisTime for inflation.

HackFisher:
It is included in the genesis config file as part of runtime.

PR #280, opened 2020-02-21
fix: #171

AurevoirXavier:


bin/node/runtime/build.rs
```diff
@@ -1,22 +1,21 @@
...
+};
 
 use wasm_builder_runner::{build_current_project_with_rustflags, WasmBuilderSource};
 
 fn main() {
+	fs::write(
+		&Path::new(&env::var_os("OUT_DIR").unwrap()).join("timestamp_now.rs"),
+		&format!(
+			"pub const NOW: u64 = {};",
+			SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis() as u64,
```
HackFisher: This is not deterministic, so there might be issues.

HackFisher:
If someone is compiling and run local, then the genesis time will not be Consistent

PR #293, opened 2020-02-26
Test/staking

yanganto:
- refactor and upgrade staking test case to 2.0


yanganto:
Some part of the staking test case is ready, the test cases of ci are passed.
In order to prevent further conflict, the test cases need to check are label and comment.
Please review the part of the test cases not in the comment.

Issue #285, opened 2020-02-24
TODO: Treasury support Kton Proposal


AurevoirXavier:

HackFisher:
Slash in staking module may collect some kton to treasury (or directly burned)?

If treasury requires to manage KTON, then we should support KTON in treasury too.

We can define RingProposal/Kton Proposal, or define enum value such as RingOrKtonBalance?

AurevoirXavier:
Slashed *Kton* goes to treasury.

HackFisher:
> Slashed _Kton_ goes to treasury.

Ok, then we should support proposals to spend KTON. (burn KTON is also spending it in some way)

clearloop:
Hi folks, I'll join this issue till it gets solved next days~ Just got some work about [GenePaper][1], logic and confusion of my mind occur as below.

### Where KTON from? 

KTON can **only obtained by locking RING**.

+ POS -> RING
  + States
    + Staking-lock -> KTON

### How to spend KTON?

1. Transfer to other accounts
  a. ...
2. burned by slashing

### Balances supports(in file structures)

+ `bin/node/runtime`
+ `frame/chainrelay/eth/backing`
+ `frame/stacking`
+ `frame/treasury`

## Abstract working steps

+ Add READMEs to `frame/balances`  
+ Complete tests according to `RING` and `Treasury Traits`
+ Review the reflecting from KTON to `treasury` and `?balances`

For `treasury` and `balances`, I'm not sure about the relationship inside them, so, **if `treasury` is the concept in substrate projects, and `balances` is the way we talking about `treasury` in the product?**


[1]: https://evolution.l2me.com/darwinia/Darwinia_Genepaper_EN.pdf

clearloop:
## Convert Balance

Darwinia has `KTON` and `RING`, both of them original by `Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>`, it might be an hack embedding them into one Generic Trait.

For example: 

```rust
pub struct Module<T: Trait> for enum Call where origin: T::Origin { 
    const ProposalBondMinimum: RingBalance<T> = T::ProposalBondMinimum::get();
    // ...
}
```
Constant `ProposalBondMinimum` require a specific `type`, it will be a mess using `enum`, because `Module` can be called by outside crates, once we change this to `enum`, it will be a lot of work.

### Solution

What about implementing `std::convert` for `KTON` and `RING`, this will cause less change **but** easy to be mixed up in some way.

[...]

PR #306, opened 2020-02-28
Treasury support KTON Proposal 

clearloop:
#285 

---

## Working Steps ✅

+ [x] Add `StakingBalance` for enum usage
+ [x] Complete  `storage`, `event` and `Module` methods.
+ [x] Spend Kton fund
+ [x] A better way to init  `ProposalBondMinimum`?
+ [x] Remove `StakingBalance`
+ [x] No spend no burn
+ [x] Complete tests.


frame/treasury/src/lib.rs
```diff
@@ -157,21 +167,52 @@ decl_module! {
...
+						ProposalCount::put(c + 1);
+
+						// TODO: @clearloop
+						//
+						// This line requires completing StakingBalance store
+						// <Proposals<T>>::insert(c, Proposal { proposer, value, beneficiary, bond });
+
+						Self::deposit_event(RawEvent::Proposed(c));
+					}
+				}
```
clearloop: The insert method requires complete storage in line 280.

frame/treasury/src/lib.rs
```diff
@@ -247,11 +302,19 @@ decl_storage! {
 				&<Module<T>>::account_id(),
 				T::RingCurrency::minimum_balance(),
 			);
+
+			// TODO: how to init both Ring and Kton in genesis?
+			//
+			// let _ = T::KtonCurrency::make_free_balance_be(
+			// 	&<Module<T>>::account_id(),
+			// 	T::KtonCurrency::minimum_balance(),
+			// );
```
clearloop: Methods like this doesn't have arguments to infer if we need to handle `RING` or `KTON`, this might need any flags?

AurevoirXavier: These are config, just 
```rust
let _ = T::KtonCurrency::make_free_balance_be(
 	&<Module<T>>::account_id(),
 	T::KtonCurrency::minimum_balance(),
);
```

clearloop: Get it, I'm trying to add some generic parameters for methods like `pot` now,

```rust
fn pot<T>() -> T::Balance where T: ... {

}
```

clearloop: Well, I think I probably understand how to master these naughty substrate traits now.

```rust
fn pot<C>() -> C::Balance 
where C: Currency<T::AccountId>,
{
    // ...
}
```

for tests: 

```rust
#[test]
fn gp() {
    let ring_balance = Treasury::pot::<Ring>::();
    let kton_balance = Treasury::pot::<Kton>::();
}
```

The usages of traits in substrate are quite like magic.

frame/treasury/src/lib.rs
```diff
@@ -374,34 +349,45 @@ impl<T: Trait> Module<T> {
 	}
 
 	// Spend some money!
-	fn spend_funds() {
-		let mut budget_remaining = Self::pot();
-		Self::deposit_event(RawEvent::Spending(budget_remaining));
+	fn spend_funds<C: Currency<T::AccountId>>() {
+		let mut budget_remaining = Self::pot::<T::RingCurrency>();
+		Self::deposit_event(RawEvent::Spending(StakingBalance::RingBalance(budget_remaining)));
```
clearloop: Here is the problem needs to be fixed currently. 

It's important to infer the passing `Currency` to spend different currency, but `Currency` is only a `trait` which doesn't have `Sized` and can .not make into an object(for inferring different passing type).

frame/treasury/src/lib.rs
```diff
@@ -108,14 +109,15 @@ pub trait Trait: frame_system::Trait {
...
+	type RingProposalRejection: OnUnbalanced<RingNegativeImbalance<Self>>;
+	type KtonProposalRejection: OnUnbalanced<KtonNegativeImbalance<Self>>;
 
 	/// Fraction of a proposal's value that should be bonded in order to place the proposal.
 	/// An accepted proposal gets these back. A rejected proposal does not.
 	type ProposalBond: Get<Permill>;
 
 	/// Minimum amount of funds that should be placed in a deposit for making a proposal.
-	type ProposalBondMinimum: Get<RingBalance<Self>>;
+	type ProposalBondMinimum: Get<StakingBalanceT<Self>>;
```
clearloop: Maybe need a better way to declare `PropoosalBondMinimum`?...haven't figure out the logic between `ExistentialDeposit`, `Get`, and `parameter_types` currently now, but it seems the `Get<StakingBalanceT<Self>>` way is not correct.

clearloop: This requires to modify the source code of [`impl<T: Trait> Currency<T::AccountId> for Module<T>`][1]

[1]: https://github.com/darwinia-network/darwinia/blob/a61fbb51b3538ca19aa8d34c528ad0f3c2c3e1ce/frame/balances/kton/src/lib.rs#L479

frame/treasury/src/lib.rs
clearloop: Solved this by a common `enum` parameter...

frame/treasury/src/lib.rs
```diff
@@ -514,14 +674,20 @@ mod tests {
 		new_test_ext().execute_with(|| {
 			// Check that accumulate works when we have Some value in Dummy already.
 			Ring::make_free_balance_be(&Treasury::account_id(), 101);
-			assert_eq!(Treasury::pot(), 100);
+			assert_eq!(Treasury::pot::<Ring>(), 100);
+			// assert_eq!(Treasury::pot::<Kton>(), 0);
```
clearloop: Not sure about why `KTON` and `RING` point to one `Currency`.

clearloop: Solved this with different `storage` names, for example:

### KTON
```rust
decl_storage! {
    trait Store for Module<T: Trait> as KtonBalances { 
        // ...
    }
}
```

### RING
```rust
decl_storage! {
    trait Store for Module<T: Trait> as RingBalances { 
        // ...
    }
}
```

frame/treasury/src/lib.rs
```diff
@@ -210,10 +260,12 @@ decl_module! {
 			Approvals::mutate(|v| v.push(proposal_id));
 		}
 
+		/// This function will implement the `OnFinalize` trait
 		fn on_finalize(n: T::BlockNumber) {
 			// Check to see if we should spend some funds!
 			if (n % T::SpendPeriod::get()).is_zero() {
-				Self::spend_funds();
+				Self::spend_funds(Funds::Ring);
+				Self::spend_funds(Funds::Kton);
```
HackFisher: The Funds enums seem to only be used here? If so, let's keep it simple by just using 

```
Self::spend_ring_funds();
Self::spend_kton_funds();
```

?

frame/treasury/src/lib.rs
```diff
@@ -299,39 +357,126 @@ impl<T: Trait> Module<T> {
...
+				StakingBalance::ring(T::RingProposalBondMinimum::get().max(T::ProposalBond::get() * value))
+			}
+		}
 	}
 
-	// Spend some money!
-	fn spend_funds() {
-		let mut budget_remaining = Self::pot();
-		Self::deposit_event(RawEvent::Spending(budget_remaining));
+	/// TODO: Need a pre-logic to infer what specific to do with kton or ring.
```
HackFisher: The budgets are from different pots, and require checking.

It is possible to reuse the code and merge spend_kton_funds and spend_ring_funds

frame/treasury/src/lib.rs
```diff
@@ -83,18 +85,17 @@ use frame_system::{self as system, ensure_signed};
 use serde::{Deserialize, Serialize};
 use sp_runtime::{
 	traits::{AccountIdConversion, EnsureOrigin, Saturating, StaticLookup, Zero},
-	ModuleId, Permill,
+	ModuleId, Permill, RuntimeDebug,
 };
 use sp_std::prelude::*;
-
-use darwinia_support::OnUnbalancedKton;
```
AurevoirXavier: import order:

```
use std...

ues third-party..

use custom..
```

bin/node/runtime/src/lib.rs
```diff
@@ -501,9 +501,11 @@ impl pallet_treasury::Trait for Runtime {
 	type ApproveOrigin = pallet_collective::EnsureMembers<_4, AccountId, CouncilCollective>;
 	type RejectOrigin = pallet_collective::EnsureMembers<_2, AccountId, CouncilCollective>;
 	type Event = Event;
-	type ProposalRejection = ();
+	type KtonProposalRejection = ();
+	type RingProposalRejection = ();
 	type ProposalBond = ProposalBond;
-	type ProposalBondMinimum = ProposalBondMinimum;
+	type RingProposalBondMinimum = ();
+	type KtonProposalBondMinimum = ();
```
AurevoirXavier:
```
type KtonProposalBondMinimum = ();
```

?

frame/treasury/src/lib.rs
```diff
@@ -222,11 +274,11 @@ decl_module! {
 /// A spending proposal.
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Encode, Decode, Clone, PartialEq, Eq, sp_runtime::RuntimeDebug)]
-pub struct Proposal<AccountId, RingBalance> {
+pub struct Proposal<AccountId, StakingBalance> {
```
AurevoirXavier: I think Proposal can be something like Ledger @HackFisher 

```rust
pub struct Proposal<AccountId, RingBalance, KtonBalance> {
	proposer: AccountId,
	proposing_ring: RingBalance,
	proposing_kton: KtonBalance,
	bond_ring: RingBalance,
	bond_kton: KtonBalance,
}
```

Also good for merge `spend_kton_funds` and `spend_ring_funds`.

HackFisher: It is a good idea, can help support more powerful combined proposal which can use RING and KTON the same time.
```
pub struct Proposal<AccountId, RingBalance> {
	proposer: AccountId,
        beneficiary: AccountId,
	ringValue: RingBalance,
	ringBond: RingBalance,
         ktonValue: KtonBalance,
	ktonBond: KtonBalance,
}
```

HackFisher: And in this way, we may remove ```StakingBalance ``` enum, it seems not required anymore.

HackFisher: For checking, ringBond must be larger than minimum, either, ktonBond must be larger than minimum. That is, it can be only proposal for ring or for kton, if the other one is zero.

frame/treasury/src/lib.rs
```diff
@@ -54,25 +54,27 @@
 //! ## GenesisConfig
 //!
 //! The Treasury module depends on the [`GenesisConfig`](./struct.GenesisConfig.html).
-
 #![cfg_attr(not(feature = "std"), no_std)]
-
+#[cfg(test)]
+mod tests;
```
clearloop: Should I move `tests` back to `lib.rs`? Move it outside is not friendly for reviewing this pr.

frame/treasury/src/lib.rs
clearloop: I don't agree with removing `StakingBalance` totally, because, in some logics, it simplify methods` logic practically.

I'll push a new version about keeping both `StakingBalance` and `Proposal<AccountId, RingBalance, KtonBalance>`, try to simplify the code as possible as it can be.

bin/node/runtime/src/lib.rs
clearloop: Add this line before because the outdated `StakingBalances` enum can not be `Get` from `u32`, now it's possible to use one `ProposalBondMinimum` for both `Ring` and `Kton`, but, independent `ProposalBondMinimum` might support more needs?

frame/treasury/src/lib.rs
```diff
@@ -299,49 +324,80 @@ impl<T: Trait> Module<T> {
...
+				if p.kton_value > budget_remaining_kton || p.kton_value == KtonBalance::<T>::from(0) {
+					should_burn_kton = false;
+					if should_return {
+						return true;
 					}
 				} else {
-					false
+					budget_remaining_kton -= p.kton_value;
+					let _ = T::KtonCurrency::unreserve(&p.proposer, p.kton_bond);
+					imbalance_kton.subsume(T::KtonCurrency::deposit_creating(&p.beneficiary, p.kton_value));
```
HackFisher: If not missing any, then will enter the burn process.
If budget is not enough, then should not be retained.

For here budget not enough, I think the requirement should be ``` p.ring_value > budget_remaining_ring || p.kton_value > budget_remaining_kton```, so it is different with the logic in the implementation.

frame/treasury/src/lib.rs
```diff
@@ -299,49 +324,80 @@ impl<T: Trait> Module<T> {
...
-						// return their deposit.
-						let _ = T::RingCurrency::unreserve(&p.proposer, p.bond);
-
-						// provide the allocation.
-						imbalance.subsume(T::RingCurrency::deposit_creating(&p.beneficiary, p.value));
-
-						Self::deposit_event(RawEvent::Awarded(index, p.value, p.beneficiary));
-						false
-					} else {
-						missed_any = true;
```
HackFisher: If ring_value > 0 then missed_any_ring = true;

if kton_value >0 then missed_any_kton = ture;

bin/node/runtime/src/lib.rs
clearloop: Changed these to constant in the lastest commit.

frame/treasury/src/lib.rs
clearloop: Fixed.

frame/treasury/src/lib.rs
```diff
@@ -299,49 +324,92 @@ impl<T: Trait> Module<T> {
...
+
+				if p.kton_value > KtonBalance::<T>::from(0) {
+					budget_remaining_kton -= p.kton_value;
+					let _ = T::KtonCurrency::unreserve(&p.proposer, p.kton_bond);
+					imbalance_kton.subsume(T::KtonCurrency::deposit_creating(&p.beneficiary, p.kton_value));
+				}
+
+				<Proposals<T>>::remove(index);
+				Self::deposit_event(RawEvent::Awarded(index, p.ring_value, p.kton_value, p.beneficiary));
+				false
```
HackFisher: Logic issues here.

frame/treasury/src/lib.rs
```diff
@@ -371,25 +371,20 @@ impl<T: Trait> Module<T> {
...
 
-				// no spent no burn
-				if p.ring_value == RingBalance::<T>::from(0) {
-					miss_any_ring = true;
-				}
-
-				if p.kton_value == KtonBalance::<T>::from(0) {
-					miss_any_kton = true;
-				}
-
```
clearloop: Should infer if balance is zero, because a possible situation is, one of kton or ring is zero.

frame/treasury/src/lib.rs
```diff
@@ -371,25 +371,20 @@ impl<T: Trait> Module<T> {
...
+				} else {
+					miss_any_ring = true;
 				}
 
 				if p.kton_value > KtonBalance::<T>::from(0) {
 					budget_remaining_kton -= p.kton_value;
 					let _ = T::KtonCurrency::unreserve(&p.proposer, p.kton_bond);
 					imbalance_kton.subsume(T::KtonCurrency::deposit_creating(&p.beneficiary, p.kton_value));
+				} else {
+					miss_any_kton = true;
```
clearloop: Merge `if-else` closures here, or maybe `==` is more efficient than `>`?

clearloop: Or if spent balances less than zero? I think I'd better change this to `!=`

frame/treasury/src/lib.rs
```diff
@@ -299,49 +324,87 @@ impl<T: Trait> Module<T> {
...
+
+					if p.kton_value > budget_remaining_kton {
+						miss_any_kton = true;
+					}
+
+					return true;
+				}
+
+				if p.ring_value > RingBalance::<T>::from(0) {
+					budget_remaining_ring -= p.ring_value;
```
HackFisher: budget_remaining_ring might < p.ring_value, overflow.

The condition checking is wrong too, ```if p.ring_value <= budget_remaining_ring && p.kton_value <= budget_remaining_kton```

frame/treasury/src/lib.rs
```diff
@@ -299,49 +324,87 @@ impl<T: Trait> Module<T> {
...
+					let _ = T::RingCurrency::unreserve(&p.proposer, p.ring_bond);
+					imbalance_ring.subsume(T::RingCurrency::deposit_creating(&p.beneficiary, p.ring_value));
+				} else {
+					miss_any_ring = true;
+				}
+
+				if p.kton_value > KtonBalance::<T>::from(0) {
```
HackFisher: Same here.

frame/treasury/src/lib.rs
```diff
@@ -299,49 +324,87 @@ impl<T: Trait> Module<T> {
...
+				// Should always be some, but shouldn't panic if false or we're screwed.
+				let option_proposal = Self::proposals(index);
+				if option_proposal.is_none() {
+					return false;
+				}
+
+				let p = option_proposal.unwrap();
+				if p.ring_value > budget_remaining_ring {
+					if p.ring_value > budget_remaining_ring {
+						miss_any_ring = true;
```
HackFisher: Duplicate checking.

frame/treasury/src/lib.rs
HackFisher: > budget_remaining_ring might < p.ring_value, overflow.
> 
> The condition checking is wrong too, `if p.ring_value <= budget_remaining_ring && p.kton_value <= budget_remaining_kton`

For here, this is impossible, so this comment can be ignored.

But for the following KTON checking, there are checking issues.

PR #307, opened 2020-02-29
Refactoring Balances

AurevoirXavier:


frame/balances/ring/src/lib.rs
```diff
@@ -897,266 +847,349 @@ where
...
-		let now = <system::Module<T>>::block_number();
-		if locks
-			.into_iter()
-			.all(|l| l.withdraw_lock.can_withdraw(now, new_balance) || !l.reasons.intersects(reasons))
-		{
-			Ok(())
-		} else {
-			Err(<Error<T, I>>::LiquidityRestrictions.into())
-		}
+		let min_balance = Self::account(who).frozen(reasons.into());
```
HackFisher: Didn't check withdraw_lockes?

frame/support/src/lib.rs
```diff
@@ -10,16 +10,75 @@ pub mod structs {
...
+			}
+		}
+
+		#[inline]
+		pub fn update_locks(&mut self, at: Moment) -> Balance {
+			match &mut self.withdraw_lock {
+				WithdrawLock::Normal(_) => Zero::zero(),
+				WithdrawLock::WithStaking(lock) => lock.update_locks(at),
+			}
+		}
```
HackFisher: Didn't get why we need update_locks here.

frame/balances/ring/src/lib.rs
```diff
@@ -1184,58 +1224,44 @@ where
...
-	type Moment = T::BlockNumber;
-
-	/// Get the amount that is currently being vested and cannot be transferred out of this account.
-	fn vesting_balance(who: &T::AccountId) -> T::Balance {
-		if let Some(v) = Self::vesting(who) {
-			Self::free_balance(who).min(v.locked_at(<system::Module<T>>::block_number()))
-		} else {
-			Zero::zero()
-		}
+		Self::update_locks(who, locks);
```
HackFisher: Is it ok no doing
```
<Locks<T, I>>::insert(who, locks);
```

frame/balances/ring/src/lib.rs
```diff
@@ -1184,58 +1224,44 @@ where
...
+	// 	Self::update_locks(who, &locks[..]);
+	// }
 
-	/// Remove a vesting schedule for a given account.
-	fn remove_vesting_schedule(who: &T::AccountId) {
-		<Vesting<T, I>>::remove(who);
+	fn remove_lock(id: LockIdentifier, who: &T::AccountId) {
+		let mut locks = Self::locks(who);
+		locks.retain(|l| l.id != id);
+		Self::update_locks(who, locks);
```
HackFisher: Why we need update_locks here?

frame/support/src/lib.rs
```diff
@@ -65,32 +142,46 @@ pub mod structs {
...
+			})
+		}
+
 		#[inline]
 		fn can_withdraw(&self, at: Moment, new_balance: Balance) -> bool {
+			new_balance >= self.locked_amount(at)
+		}
+
+		#[inline]
+		fn update_locks(&mut self, at: Moment) -> Balance {
```
HackFisher: This is only called when set_lock(), suggest to move to staking model, to keep lock logic simple.

frame/balances/ring/src/lib.rs
AurevoirXavier: Keep origin logic with Substrate

frame/balances/ring/src/lib.rs
AurevoirXavier: `update_locks` will call `insert`.

frame/balances/ring/src/lib.rs
```diff
@@ -530,122 +477,137 @@ decl_module! {
...
-		// assert!(Self::free_balance(who) > T::ExistentialDeposit::get());
-		if balance < T::ExistentialDeposit::get() {
-			<FreeBalance<T, I>>::insert(who, balance);
-			Self::on_free_too_low(who);
-			UpdateBalanceOutcome::AccountKilled
-		} else {
-			<FreeBalance<T, I>>::insert(who, balance);
-			UpdateBalanceOutcome::Updated
+	/// Get the frozen balance of an account.
+	fn frozen_balance(who: impl Borrow<T::AccountId>) -> FrozenBalance<T::Balance> {
```
HackFisher: suggest to renaming to ```extended_locked_balance``` to avoid terminology confuse, because there is another ```frozen(``` method in FrozenBalance.

AurevoirXavier: `frozen_balance` looks good. `frozen` need a reason param, change the `frozen` to `frozen_for`?

AurevoirXavier:
**Can not Compile**. Require #310 Full Upgrade.

Issue #325, opened 2020-03-09
Keep test case and ci updated when developing


yanganto:
The test and ci are also parts of the development.
There is no KTON test case in the current `develop` branch and the Staking test case is reverted.
- The `.maintain/ci/darwinia_test_script.sh` should be updated with #307 merged.
- The staking test case should be kept in passed without comment when #312 merged, this also revert the work of #293.

clearloop:
Well, I'm trying to fix `treasury`'s tests now.

AurevoirXavier:
We don’t need the KTON tests. Those two module use the same code.

clearloop:
> We don’t need the KTON tests. Those two modules use the same code.

I realized it this afternoon, I'll keep some functional tests with KTON parallel in treasury's tests, and delete most of them.

Issue #320, opened 2020-03-06
Merge Ring and Kton


AurevoirXavier:
https://github.com/paritytech/substrate/issues/5175

HackFisher:
we may achieve this by supporting two kinds of AccountData, etc.

and a trait for balance to retrieving data.

free<Balance>

reserve<Balance>
```
AccountDataX<BalanceRing, BalanceKton> {
pub free_ring: BalanceRing,
pub free_kton: BalanceKton,
pub reserved_ring: BalanceRing,
pub reserved_kton: BalanceKton,

free<BalanceRing>: {
return free_ring;
}
}
```

or 

```
AccountDatay<BalanceA> {
pub free: BalanceA,
pub reserved: BalanceA,
}
```

PR #326, opened 2020-03-09
Fix treasury's tests for v2.0.0-alpha.3

clearloop:
#325 

---

## Steps

+ [x] Migrate new tests from v2.0.0-alpha.3
+ [x] Fix tests

### DIFF
--- a/frame/treasury/src/mock.rs
+++ b/frame/treasury/src/mock.rs
@@ -116,7 +116,6 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
 	}
 	.assimilate_storage(&mut t)
 	.unwrap();
-
 	pallet_kton::GenesisConfig::<Test> {
 		// Total issuance will be 200 with treasury account initialized at ED.
 		balances: vec![(0, 100), (1, 98), (2, 1)],
--- a/frame/treasury/src/mock.rs
+++ b/frame/treasury/src/mock.rs
@@ -116,12 +116,13 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
 	}
 	.assimilate_storage(&mut t)
 	.unwrap();
-	// pallet_kton::GenesisConfig::<Test> {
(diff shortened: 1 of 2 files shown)


frame/treasury/src/mock.rs
```diff
@@ -116,12 +116,12 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
...
-		balances: vec![(0, 100), (1, 98), (2, 1)],
-	}
-	.assimilate_storage(&mut t)
-	.unwrap();
+	// pallet_kton::GenesisConfig::<Test> {
+	// 	// Total issuance will be 200 with treasury account initialized at ED.
+	// 	balances: vec![(0, 100), (1, 98), (2, 1)],
+	// }
+	// .assimilate_storage(&mut t)
+	// .unwrap();
```
clearloop: KTON's GenesisConfig will override the RING's, haven't figured out why yet.

AurevoirXavier: I'll make a pr to fix this

clearloop:
Should support `tip` for KTON? If so, it might be an ugly support...or we support it after we merge KTON and RING?

AurevoirXavier:
> Should we support `tip` for KTON? If so, it might be an ugly support...or we support it after we merge KTON and RING?

The merge means avoid the dup code, they still there. You probably still need two types to acess the different storage.

HackFisher:
> Should we support `tip` for KTON? If so, it might be quite ugly currently now...or we support it after we merge KTON and RING?
> 
> #320

Let's do it later or in a separate issue.

AurevoirXavier:
> > Should we support `tip` for KTON? If so, it might be quite ugly currently now...or we support it after we merge KTON and RING?
> > #320
> 
> Let's do it later or in a separate issue.

Yes.

And I think the staking tests (#300) are the highest priority.
You can pick it up and work with @yanganto .

clearloop:
Ok, I'll join #300 after merging this pr.

PR #328, opened 2020-03-09
fix: genesis for ring and kton

AurevoirXavier:


AurevoirXavier:
@clearloop https://github.com/darwinia-network/darwinia/pull/326#discussion_r389596976

Issue #289, opened 2020-02-25
Offchain worker


yanganto:
use an off-chain worker instead of darwinia-proof-tools

PR #335, opened 2020-03-12
Add ethereum offchain worker

yanganto:
- create an off-chain worker to submit the ether block headers
- #289


bin/node/runtime/src/lib.rs
```diff
@@ -91,6 +92,26 @@ pub type DealWithFees = SplitTwoWays<
 	Author, // 1 part (20%) goes to the block author.
 >;
 
+type SubmitPFTransaction =
+	frame_system::offchain::TransactionSubmitter<pallet_eth_offchain::crypto::Public, Runtime, UncheckedExtrinsic>;
+
+parameter_types! {
+   pub const BlockFetchDur: BlockNumber = 3;
+   // TODO: pass this from command line
+   pub const EthScanAPIKey: [u8; 34] = [74 as u8, 53 as u8, 51 as u8, 55 as u8, 68 as u8, 68 as u8,
```
HackFisher: https://substrate.dev/docs/en/development/module/off-chain-workers#option-2-add-an-app-subkey-via-cli

frame/chainrelay/eth/offchain/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+		for _ in 0..(width - content.len() + 2) {
+			output.push(48);
+		}
+		output.append(&mut content.into_bytes()[2..].to_vec());
+		return output.iter().map(|u| *u as char).collect::<AllocString>();
+	}
+	content
+}
+
+fn build_eth_header(number: u64, block_info: JsonValue) -> Result<EthHeader> {
```
HackFisher: How about build a off-chain api wrap service, and do the parse stuffs off chain?

bin/node/runtime/src/lib.rs
yanganto: Thanks, I will find the solution from this document.

frame/chainrelay/eth/offchain/src/lib.rs
yanganto: This worker gets the block info now relies on Ethscan's API, so it already wraps a web service.  

I think the way to avoid parsing a string into bytes is to use more low-level client,  and there is still no grpc/socket client API in the off-chain worker.  So I think parsing is not avoidable.

By the way, the response format of Ethscan is a JSON with sorted fields, and also is the same as the Ethereum gateway of Cloudflare.  And the worker is easily integrated with multiple sources and will be implemented later to enhance the robust.

AurevoirXavier:
If I don't use a VPN.

```rust
WARN wasm-runtime  Ran out of free WASM instances
```

yanganto:
> If I don't use a VPN.
> 
> ```rust
> WARN wasm-runtime  Ran out of free WASM instances
> ```

This issue may not exist, when the #370 implemented and removed the retry feature.

Issue #265, opened 2020-02-14
Follow Substrate CI


AurevoirXavier:
**https://github.com/paritytech/substrate/blob/master/.gitlab-ci.yml**

---

**Cache Here**
```yml
variables:
  GIT_STRATEGY:                    fetch
  GIT_DEPTH:                       100
  CARGO_HOME:                      "/ci-cache/${CI_PROJECT_NAME}/cargo/${CI_JOB_NAME}"
  SCCACHE_DIR:                     "/ci-cache/${CI_PROJECT_NAME}/sccache"
  CARGO_INCREMENTAL:               0
  CI_SERVER_NAME:                  "GitLab CI"
  DOCKER_OS:                       "debian:stretch"
  ARCH:                            "x86_64"
```

yanganto:
Thanks.

yanganto:
There may be a bug when storing the cache in Travis ci.  The storing cache may cause the job to fail by reaching the time limit.  If the issue can not be solved, we will use GitLab ci.

yanganto:
Some notes about this issue.
https://www.reddit.com/r/rust/comments/dbgril/caching_on_travis_when_building_rust_projects/

yanganto:
The GitLab of parity is self-hosted.
`GITLAB_API:                    "https://gitlab.parity.io/api/v4"`

yanganto:
Pipline artifacts save times
https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml

PR #341, opened 2020-03-13
Add Circle CI

yanganto:
- move part of ci to Circle ci to save time
- use cache artifacts from the build job
- #265


yanganto:
Postponed and will reopen after Crab net launch

PR #343, opened 2020-03-13
Review

AurevoirXavier:


frame/chainrelay/eth/relay/src/lib.rs
```diff
@@ -296,64 +291,68 @@ impl<T: Trait> Module<T> {
...
+			best_header_info.number <= header.number + Self::number_of_blocks_finality(),
+			"Header Too Old: It's too late to add this block header.",
+		);
 
 		let header_hash = header.hash();
+		// FIXME check write storage
 		HeaderOf::insert(header_hash, header);
 
-		let parent_total_difficulty = Self::header_info_of(header.parent_hash())
+		let parent_total_difficulty = Self::header_info_of(header.parent_hash)
```
HackFisher: Side effect on Error

frame/chainrelay/eth/relay/src/lib.rs
```diff
@@ -325,9 +377,10 @@ impl<T: Trait> Module<T> {
 
 			// If the new header has a lower number than the previous header, we need to cleaning
 			// it going forward.
-			// FIXME check overflow
 			if best_header_info.number > header_info.number {
-				for number in header_info.number + 1..=best_header_info.number {
+				for number in
```
HackFisher: It's ok to use check_add here, but can also omit overflow check after carefully review.(Can't overflow here)

Issue #123, opened 2019-11-28
Review: Weight for Extrinsics


HackFisher:

clearloop:
## [Transaction Weight][1]

A number of resources in any chain can be limited, such as storage or computation. Weights exist to prevent individual components of the chain from consuming too much of any resource.

Consuming some weights should generally incur a fee. The fee implications of the weight system are covered in the [Fee Developer document][2].


## Weight in current Darwinia

+ Transaction
  + [RING][3] and [KTON][4] balances
  + [Treasury][5]
+ Nomination
  + [Nomination in staking][6]
+ Voting
  + [Voting in elections-phragmen][7]

## References

+ [Transaction Weight   (substrate doc)][1]
+ [Weight system in substrate   (substrate doc)][2]
+ [Calculate weight in substrate    (ZH substrate guid)][8]
+ [`estimate_fee` in web3    (web3 doc)][9]
+ [Transaction fee estimation    (substrate issue)][10]

[1]: https://substrate.dev/docs/en/next/conceptual/runtime/weight
[2]: https://substrate.dev/docs/en/next/development/module/fees
[3]: https://github.com/darwinia-network/darwinia/blob/bee19e38baaa62cb529fd839e16585bfe2340c43/frame/balances/ring/src/lib.rs#L363
[4]: https://github.com/darwinia-network/darwinia/blob/bee19e38baaa62cb529fd839e16585bfe2340c43/frame/balances/kton/src/lib.rs#L363
[5]: https://github.com/darwinia-network/darwinia/blob/bee19e38baaa62cb529fd839e16585bfe2340c43/frame/treasury/src/lib.rs#L135
[6]: https://github.com/darwinia-network/darwinia/blob/bee19e38baaa62cb529fd839e16585bfe2340c43/frame/staking/src/lib.rs#L74
[7]: https://github.com/darwinia-network/darwinia/blob/bee19e38baaa62cb529fd839e16585bfe2340c43/frame/elections-phragmen/src/lib.rs#L33
[8]: https://zhuanlan.zhihu.com/p/109696123?utm_source=wechat_session&utm_medium=social&utm_oi=52957873700864&s_s_i=oAd663Y%2F2YAtuV1kkjLQfZ0rqIrqYmI1JIQPvIPpgOU%3D&s_r=1&from=groupmessage&isappinstalled=0
[9]: https://web3js.readthedocs.io/en/v1.2.0/web3-eth-contract.html#methods-mymethod-estimategas
[...]

PR #354, opened 2020-03-17
Review all weights

clearloop:
#123 

## Steps

+ [x] balances
+ [x] treasury
+ [x] chainrelay
  + [x] relay
  + [x] baking
+ [x] staking
+ [x] elections-phgramen

### DIFF
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1392,9 +1392,6 @@ decl_module! {
 			}
 		}
 
-		/// Claim mature deposits from a controller account back to the related
-		/// stash accounts.
-		///
 		/// Stash accounts can get their ring back after the depositing time exceeded,
 		/// and the ring getting back is still in staking status.
 		///
--- a/frame/staking/src/lib.rs
+++ b/frame/staking/src/lib.rs
@@ -1229,7 +1229,16 @@ decl_module! {
 			}
 		}
 
-		// TODO: doc
+		/// Deposit some extra amount ring, and return kton to the controller.
+		///
+		/// The dispatch origin for this call must be _Signed_ by the stash, not the controller.
+		///
+		/// # <weight>
+		/// - Independent of the arguments. Insignificant complexity.
+		/// - O(1).
+		/// - One DB entry.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
 		fn deposit_extra(origin, value: RingBalance<T>, promise_month: Moment) {
 			let stash = ensure_signed(origin)?;
 			let controller = Self::bonded(&stash).ok_or(<Error<T>>::NotStash)?;
@@ -1287,7 +1296,7 @@ decl_module! {
 		/// </weight>
 		///
 		/// Only active normal ring can be unbond
-		#[weight = SimpleDispatchInfo::FixedNormal(400_000)]
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
 		fn unbond(origin, value: StakingBalanceT<T>) {
 			let controller = ensure_signed(origin)?;
 			let mut ledger = Self::clear_mature_deposits(Self::ledger(&controller).ok_or(<Error<T>>::NotController)?);
@@ -1383,7 +1392,15 @@ decl_module! {
 			}
 		}
 
-		// TODO: doc
+		/// Stash Account can get their ring back after the depositing time exceeded.
+		///
+		/// # <weight>
+		/// - Independent of the arguments. Insignificant complexity.
+		/// - One storage read.
+		/// - One storage write.
+		/// - Writes are limited to the `origin` account key.
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
 		fn claim_mature_deposits(origin) {
 			let controller = ensure_signed(origin)?;
 			let ledger = Self::clear_mature_deposits(Self::ledger(&controller).ok_or(<Error<T>>::NotController)?);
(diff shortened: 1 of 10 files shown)


frame/staking/src/lib.rs
```diff
@@ -1755,6 +1781,13 @@ decl_module! {
 		/// This can be called from any origin.
 		///
 		/// - `stash`: The stash account to reap. Its balance must be zero.
+		///
+		/// # <weight>
+		/// - `O(1)`
+		/// - One slash kill
+		/// - Two lock removes
+		/// # </weight>
+		#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
```
HackFisher: I have question that why original didn't require weight? Will unsigned extrinsics successful?

Need someone else to review this.

frame/staking/src/lib.rs
```diff
@@ -1541,7 +1567,7 @@ decl_module! {
 		/// Effects will be felt at the beginning of the next era.
 		///
 		/// The dispatch origin for this call must be _Signed_ by the controller, not the stash.
-		///
+		///1
```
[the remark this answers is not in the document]
clearloop: Yep, I'm waiting for the `travis` and thinking about if I need raise another force push...

frame/staking/src/lib.rs
clearloop: I'm not sure about this, so applied the lowest weight in darwinia weight standard, but this method seems not completed yet.

```rust
#[weight = SimpleDispatchInfo::FixedOperational(10_000)]
fn reap_stash(_origin, stash: T::AccountId) {
    Self::ensure_storage_upgraded();
    ensure!(T::RingCurrency::total_balance(&stash).is_zero(), <Error<T>>::FundedTarget);
    ensure!(T::KtonCurrency::total_balance(&stash).is_zero(), <Error<T>>::FundedTarget);

    Self::kill_stash(&stash)?;
    T::RingCurrency::remove_lock(STAKING_ID, &stash);
    T::KtonCurrency::remove_lock(STAKING_ID, &stash);
 }
```

However, the `origin` is `_origin` and doesn't have a `ensure!`, so the system might not know how to spend weights, it might be safer to keep no weight costs as the original implementation.

frame/staking/src/lib.rs
```diff
@@ -1383,15 +1392,32 @@ decl_module! {
...
+		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
 		fn claim_mature_deposits(origin) {
 			let controller = ensure_signed(origin)?;
 			let ledger = Self::clear_mature_deposits(Self::ledger(&controller).ok_or(<Error<T>>::NotController)?);
 
 			<Ledger<T>>::insert(controller, ledger);
 		}
 
-		// TODO: doc
+		/// Claim deposits when the depositing time has not been exceeded, accounts' ring
```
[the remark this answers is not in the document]
clearloop: I'll migrate the doc inside above link to this method.

frame/staking/src/lib.rs
```diff
@@ -1383,15 +1392,32 @@ decl_module! {
 			}
 		}
 
-		// TODO: doc
+		/// Stash Account can get their ring back after the depositing time exceeded.
```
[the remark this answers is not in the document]
clearloop: Still in staking status(as a ledger), and back to accounts if signer unbond them?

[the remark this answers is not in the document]
HackFisher: Controller and stash are just different keys of the same staker account.

Please read: https://wiki.polkadot.network/docs/en/learn-keys#controller-and-stash-keys

Following is ok:
> Stash accounts can get their ring back after the depositing time exceeded, the ring
> getting back is still in staking status.

frame/staking/src/lib.rs
```diff
@@ -1383,15 +1392,38 @@ decl_module! {
 			}
 		}
 
-		// TODO: doc
+		/// Claim mature deposits from a controller account back to the related
+		/// stash accounts.
+		///
```
HackFisher: Remove the three lines.

frame/staking/src/lib.rs
HackFisher: How will the system calculate the fees if we don't set the weight tags?

HackFisher: The default is ```SimpleDispatchInfo::FixedNormal(10_000)```

Refer https://zhuanlan.zhihu.com/p/108194544 and the code: 
https://github.com/paritytech/substrate/blob/master/frame/support/src/weights.rs#L233

> 在有限的区块生成时间和链上状态的限制下，权重被用来定义交易产生的计算复杂度即所消耗的计算资源，以及占据的链上状态。system模块定义了区块的总权重（MaximumBlockWeight）。为了保证在网络繁忙的情况下，依然能够实现对区块链应用有效合理的管理，Substrate引入了两种不同级别的交易类型，既 Normal 和 Operational。Normal类型的交易是由网络中的普通用户提交，Operational类型的交易是由网络中的管理员或者管理委员会共同触发。区块资源如长度和总权重按照一定比例在这两种类型的交易中进行分配，这一比例称为可用区块比（AvailableBlockRatio）。Kusama网络的设置为：
> 
> 区块的总权重：1,000,000,000
> 可用区块比：75%，即Normal交易最多只占用75%的区块资源，Operational类型的交易则可以占用100%的区块资源，新的交易如果导致对应资源使用率超过阈值后，会被拒绝。

FixNormal and Operational is for different purpose, so we should not use Operational here.

10_000 or default (not set) is both OK.

clearloop: Get it, I'll apply these explainations to the-darwinia-book's weight chapter.

PR #358, opened 2020-03-18
Support MMR digest in block header

### LINKED ISSUE #344: Support and include MMR in blockheader
For easier darwinia relay support

### LINKED ISSUE #352: Make MMR Into a Module
Which can deposit log with its own module Id.

HackFisher:
Closes #352 
Closes #344 

This is to support and include MMR in darwinia block header, so the darwinia's light client and chain relay can use this mmr value to build super light client.

Blockchain history digest item that contains the merkle mountain range root at given block. It is created for providing super light client a commitment of all the previous blocks.

+ [x] Introduce mmr lib and update_logs in header (System::on_finalize)
+ [x] Update mmr lib refer:
    - https://github.com/tari-project/tari/tree/development/base_layer/mmr
    - https://github.com/nervosnetwork/merkle-mountain-range
+ [x] Replace ```blake2``` with ```T::Hashing::hash_of```
+ [x] Refactor concern: Move deposit digest log from System to Darwinia's Module
+ [x] Does it require special handling in Genesis block? (What's genesis block's parent hash)
+ [x] Won't support [a. merge hash with index?](https://github.com/mimblewimble/grin/blob/0ff6763ee64e5a14e70ddd4642b99789a1648a32/core/src/ser.rs#L573)  [b. detail in push](https://github.com/mimblewimble/grin/blob/0ff6763ee64e5a14e70ddd4642b99789a1648a32/core/src/core/pmmr.rs#L279)    [c. with total difficulty](https://github.com/nervosnetwork/merkle-mountain-range/blob/master/src/tests/test_accumulate_headers.rs#L38)
+ [x] Get Proof RPC Call and verify function for testing. (verify rpc might not need since it should happen on other chains.)
+ [x] Tests
+ [x] Review append only ModuleMMRStore and checking the logic
+ [x] Document
+ [x] Update substrate dependency


frame/header-mmr/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+//use codec::{Encode, Decode};
+use sp_runtime::{
+	generic::DigestItem,
+	traits::{Hash, One},
+	DispatchError,
+};
+
+use merkle_mountain_range::{MMRStore, MerkleProof, MMR};
+
+mod mock;
```
AurevoirXavier: cfg missing

frame/header-mmr/src/lib.rs
```diff
@@ -0,0 +1,189 @@
+//! # Chain MMR Pallet
+//!
+//! \## Overview
```
AurevoirXavier: why add a `\` prefix?

HackFisher: It's copy from the sample pallet

frame/header-mmr/src/lib.rs
HackFisher: ```#![cfg(test)]```

It is defined in mock.rs and test.rs, follows the convention of the sample pallet in substrate

HackFisher: https://github.com/paritytech/substrate/blob/v2.0.0-alpha.4/frame/elections/src/mock.rs#L19

https://github.com/paritytech/substrate/blob/v2.0.0-alpha.4/frame/elections/src/lib.rs#L42

frame/header-mmr/src/lib.rs
HackFisher: https://github.com/paritytech/substrate/blob/v2.0.0-alpha.4/frame/example/src/lib.rs#L127

Should used in <pre> code snippets
Fixed.

PR #387, opened 2020-03-24
Fix ci test case flow

yanganto:
- let Darwinia test run parallel
- comment overall test (over the time limitation)


yanganto:
There are 4 workers who can run parallel in the test stage.
https://travis-ci.org/github/darwinia-network/darwinia/builds/666161190

PR #401, opened 2020-03-26
add dependency checking in precommit

yanganto:
- show the dependency check as a warning on pre-commit


AurevoirXavier:
Should move to `darwinia-common`.

yanganto:
I will copy this to that project after this PR is confirmed.

AurevoirXavier:
There's no frame or primitives in this repo.

.maintain/ensure-deps.sh
```diff
@@ -0,0 +1,80 @@
...
+
+VIOLATIONS=()
+PACKAGES=()
+
+function check_rule() {
+	rule=$1
+	from=$(echo $rule | cut -f1 -d\ )
+	to=$(echo $rule | cut -f2 -d\/)
+
+	cd $from
```
HackFisher: we have different folder structure with substrate.

so I double this will work for client, frame and primitives, and frame and primitives have already been moved to darwinia-common

.maintain/ensure-deps.sh
```diff
@@ -0,0 +1,80 @@
...
+	echo "Checking rule '$rule'"
+	packages=$(find -name Cargo.toml | xargs grep -wn "path.*\.\.\/$to")
```
HackFisher: same here, the path..

We need some customize to this script

.maintain/ensure-deps.sh
yanganto: The `$to` from `$rule` is fixed, and the `$from` part is not fixed, because we have no rule run into this at this moment.  Maybe it is a good issue for gitcoin bounty.

.maintain/ensure-deps.sh
yanganto: Under the current rule, this will be fine.

yanganto:
This PR is moved to https://github.com/darwinia-network/darwinia-common/pull/13

PR #405, opened 2020-03-31
Adjust Structure

AurevoirXavier:


AurevoirXavier:
Ready for review, but need to update the deps later.

src/main.rs
```diff
@@ -1,18 +1,21 @@
...
+
+fn main() -> Result<(), darwinia_cli::Error> {
+	let version = VersionInfo {
+		name: "Darwinia-Network Darwinia",
 		commit: env!("VERGEN_SHA_SHORT"),
 		version: env!("CARGO_PKG_VERSION"),
 		executable_name: "darwinia",
 		author: "Darwinia Network <hello@darwinia.network>",
-		description: "Darwinia node implementation in Rust",
+		description: "Darwinia Relay-chain Client Node",
```
HackFisher, line 15: We do not act as Relay Chain any more, change to "Darwinia node implementation in Rust"

src/main.rs
```diff
@@ -1,18 +1,21 @@
...
 
-fn main() -> sc_cli::Result<()> {
-	let version = sc_cli::VersionInfo {
-		name: "Darwinia Node",
+// --- darwinia ---
+use darwinia_cli::VersionInfo;
+
+fn main() -> Result<(), darwinia_cli::Error> {
+	let version = VersionInfo {
+		name: "Darwinia-Network Darwinia",
```
HackFisher, line 10: Change to "Darwinia Network"

PR #411, opened 2020-04-02
Insert token into offchain worker

yanganto:
- user can use their own token into off-chain worker
- the key can be inserted by subkey through RPC
- ref: https://github.com/darwinia-network/darwinia-common/pull/37
- ref: https://github.com/darwinia-network/substrate/pull/2


yanganto:
Due to the spec change, this PR is closed.
The detail of spec change please refer https://github.com/darwinia-network/substrate/pull/2
The another PR for the same feature #416

PR #416, opened 2020-04-08
Insert key into offchain DB

### LINKED ISSUE #391: Passing API Key into keystore for offchain worker
Currently, we use the environment variable to customize a build.   It is better to pass API Key into Keystore, such that different validator may use their own key, such that the relay will become much more robust.

yanganto:
Following frontend are suggested from [Jam](https://github.com/joshua-mir) in parity.
- https://polkadot.js.org/api/substrate/rpc.html#localstorageset-kind-storagekind-key-bytes-value-bytes-null (JS)
- https://github.com/paritytech/substrate-subxt (Rust)
- fix #391


yanganto:
dependence on https://github.com/darwinia-network/darwinia-common/pull/43

AurevoirXavier:
Move to #421

Issue #418, opened 2020-04-08
Update Crab Root Key 


HackFisher:
Current root key in crab runtime is not right, need to be updated.

PR #423, opened 2020-04-10
Add Locked Balance into Contract 

freehere107:


HackFisher:
> Change beneficiary from lock contract address to contract beneficiary, I have seen contract beneficiary address has been increased with the proper amount but didn't see entries pointing lock contract address been removed?

@freehere107 Correct me if I'm wrong, contracts holder address are all ignored in the airdrop statistic expect those related to gringott.

BTW, authorities' balances in genesis need to be update, to keep balance of the total supply, please cut them from the root key.

HackFisher:
I'll make a patch to fix the supply and root key ( Closes #418 )

PR #490, opened 2020-08-06
Release 0.6.6

AurevoirXavier:


service/src/chain_spec.rs
```diff
@@ -441,7 +440,7 @@ pub fn crab_testnet_genesis(
 				.map(|x| (x.0, x.1, 1 << 60, crab_runtime::StakerStatus::Validator))
 				.collect(),
 			invulnerables: initial_authorities.iter().cloned().map(|x| x.0).collect(),
-			force_era: crab_runtime::Forcing::NotForcing,
+			force_era: crab_runtime::Forcing::ForceAlways,
```
HackFisher, line 443: Will ```ForceAlways``` be default behaviour?

AurevoirXavier: Easy test for dev-net.

PR #522, opened 2020-09-10
Companion for #274

AurevoirXavier:
https://github.com/darwinia-network/darwinia-common/pull/274

### DIFF
--- a/runtime/crab/src/lib.rs
+++ b/runtime/crab/src/lib.rs
@@ -792,6 +792,16 @@ impl darwinia_relayer_game::Trait<EthereumRelayerGameInstance> for Runtime {
 
 impl darwinia_header_mmr::Trait for Runtime {}
 
+parameter_types! {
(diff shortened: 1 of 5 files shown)


runtime/crab/src/lib.rs
```diff
@@ -792,6 +792,16 @@ impl darwinia_relayer_game::Trait<EthereumRelayerGameInstance> for Runtime {
 
 impl darwinia_header_mmr::Trait for Runtime {}
 
+parameter_types! {
+	pub const CrabIssuingModuleId: ModuleId = ModuleId(*b"da/crabk");
```
HackFisher, line 796: "da/crais" is better than "da/crabk"

"da/crabk" could be used later on Darwinia side.

HackFisher (changes requested):

HackFisher (approved):

PR #545, opened 2020-09-21
Add Democracy

AurevoirXavier:


runtime/crab/polkadot-compatible-types.json
```diff
@@ -243,6 +243,7 @@
 			"_enum": {
 				"Any": null,
 				"NonTransfer": null,
+				"ProxyType": null,
```
AsceticBear, line 246: A bit confused:

What are `polkadot-compatible-types.json` and `type.json` used for?  They have extractly the same content.

AurevoirXavier: No, their content are different

AurevoirXavier:
Waiting https://github.com/darwinia-network/darwinia-common/pull/301

runtime/crab/polkadot-compatible-types.json
AurevoirXavier: One for darwinia apps another for polkadot apps

HackFisher (changes requested):
Need to bump Crab Runtime's Spec version

HackFisher (approved):

Issue #427, opened 2019-07-03
Load Boot Configuration from File


chenym1992:
启动配置希望可以指定配置文件

AurevoirXavier:
This feature will be delayed.

Due to https://github.com/paritytech/substrate/issues/5936 that I can't mutate the config.

AurevoirXavier:
https://github.com/paritytech/substrate/pull/5942

This will be fixed in the next substrate upgrade.

PR #594, opened 2021-01-04
fix(rpc): companion for #427

AurevoirXavier:
https://github.com/darwinia-network/darwinia-common/pull/427

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,9 +12,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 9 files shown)


HackFisher:
Since this version only changes the rpc codes, we do not to upgrade runtime, right?

AurevoirXavier:
> Since this version only changes the rpc codes, we do not to upgrade runtime, right?

Yes

HackFisher (approved):

PR #609, opened 2021-02-01
Release 0.9.3

AurevoirXavier:
Follow https://github.com/paritytech/polkadot/pull/1443

### DIFF
--- a/runtime/darwinia/src/lib.rs
+++ b/runtime/darwinia/src/lib.rs
@@ -173,13 +173,13 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("Darwinia"),
 	impl_name: create_runtime_str!("Darwinia"),
 	authoring_version: 0,
-	spec_version: 20,
+	spec_version: 21,
 	impl_version: 0,
 	#[cfg(not(feature = "disable-runtime-api"))]
 	apis: RUNTIME_API_VERSIONS,
 	#[cfg(feature = "disable-runtime-api")]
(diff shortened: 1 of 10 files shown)


runtime/darwinia/src/lib.rs
```diff
@@ -173,13 +173,13 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
...
 	authoring_version: 0,
-	spec_version: 20,
+	spec_version: 21,
 	impl_version: 0,
 	#[cfg(not(feature = "disable-runtime-api"))]
 	apis: RUNTIME_API_VERSIONS,
 	#[cfg(feature = "disable-runtime-api")]
 	apis: sp_version::create_apis_vec![[]],
-	transaction_version: 0,
+	transaction_version: 1,
```
HackFisher, line 182: When should we bump transaction_version? Any signals or convention to follow?

HackFisher:
@AurevoirXavier Since this is an important public event for community with prediction, may wish to test again on local testnet though I'm quite sure it will work.

runtime/darwinia/src/lib.rs
AurevoirXavier: If the payload changed, bump it. Else, whatever.

I would like to follow polkadot here.

AurevoirXavier:
> @AurevoirXavier Since this is an important public event for community with notice, may wish to test again on local testnet though I'm quite sure it will work.

Okay. I'm on it.

HackFisher (approved):

PR #610, opened 2021-02-02
Fix Properties

AurevoirXavier:
This change doesn't affect darwinia & crab but darwinia-dev & crab-dev

### DIFF
--- a/node/service/src/chain_spec/crab.rs
+++ b/node/service/src/chain_spec/crab.rs
@@ -45,10 +45,8 @@ pub fn crab_properties() -> Properties {
 	let mut properties = Properties::new();
 
 	properties.insert("ss58Format".into(), 42.into());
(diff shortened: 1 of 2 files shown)


AsceticBear:
> This change doesn't affect darwinia & crab but darwinia-dev & crab-dev

Why does this change not affect darwinia, crab network?  I noticed that `darwinia_development_config`, `darwinia_build_spec_config` all depend on this one configuration.

AurevoirXavier:
Because Darwinia use https://github.com/darwinia-network/darwinia/blob/master/node/service/res/darwinia/darwinia.json Crab use https://github.com/darwinia-network/darwinia/blob/master/node/service/res/darwinia/crab.json

AsceticBear (approved):

PR #620, opened 2021-02-28
doc: fix typo and improve comment statements

songtianyi:


AurevoirXavier (changes requested):
I prefer `NOT EXISTED`

songtianyi:
> I prefer `NOT EXISTED`

https://www.reddit.com/r/EnglishLearning/comments/do8mix/which_one_is_it_correct_it_is_not_existed_or_it/
https://forum.wordreference.com/threads/is-existed-vs-does-exist.2486634/

> "To exist" is an intransitive verb. You cannot exist something else. Something can only exist by itself. Therefore, it has no passive sense, so "It is not existed" cannot be correct - even though it is possible to create that form by using the standard pattern for the passive.

AurevoirXavier:
> > I prefer `NOT EXISTED`
> 
> https://www.reddit.com/r/EnglishLearning/comments/do8mix/which_one_is_it_correct_it_is_not_existed_or_it/
> https://forum.wordreference.com/threads/is-existed-vs-does-exist.2486634/
> 
> > "To exist" is an intransitive verb. You cannot exist something else. Something can only exist by itself. Therefore, it has no passive sense, so "It is not existed" cannot be correct - even though it is possible to create that form by using the standard pattern for the passive.

'had not existed' is also ok. And I want to keep the error msg [clear and short](https://github.com/darwinia-network/darwinia-common/blob/master/frame/bridge/ethereum/relay/src/lib.rs#L181)

runtime/darwinia/src/constants.rs
```diff
@@ -132,12 +132,12 @@ pub mod relay {
 
 					sample_points.push(new_sample_points);
 				} else {
-					// Should never enter this condition
-					error!(target: "ethereum-relayer-game", "Sample Round - NOT EXISTED");
+					// Should never be reached
+					error!(target: "ethereum-relayer-game", "Sample Round - DOES NOT EXIST");
```
AurevoirXavier, line 136: Also here.

songtianyi: please check again thx

AurevoirXavier:
@wi1dcard `CI: Error: Specify secrets.SLACK_WEBHOOK_URL`

AurevoirXavier (approved):

PR #683, opened 2021-07-20
CI with Glibc-2.23, Clang-4.0

AurevoirXavier:


.maintain/docker/Dockerfile.x86_64-linux-gnu
```diff
@@ -16,13 +16,15 @@
...
 # for gernal linux
 
 # change mirrorlist
-RUN yum -y update && yum -y upgrade && yum -y install \
-	# tool
-	git make \
-	# compiler
-	clang gcc gcc-c++ llvm
+RUN apt update && \
+    apt upgrade -y && \
```
wi1dcard, line 25: This can be removed I think, and the base image would be better if we change the precise version "16.04" to "xenial".

HackFisher: Will this make some os using old version such as "16.04"  unable to use the binary? @AurevoirXavier @wi1dcard

wi1dcard: Don't think so, since we're using a relatively lower version (ubuntu 16.xx) to build the binaries. However, I haven't ever tried it.

HackFisher: "xenial" means "Ubuntu 16.04 LTS"?

wi1dcard: BTW, "xenial" is pretty much equal to "16.xx".

wi1dcard: > "xenial" means "Ubuntu 16.04 LTS"?

Correct. It locks to the specific release and won't bump to the newer major versions like 18.04, 18.10, or 20.04.

HackFisher (approved):

AurevoirXavier:
@fewensa Request review

wi1dcard (approved):

fewensa (approved):

PR #686, opened 2021-07-23
Release v0.11.1

AurevoirXavier:


runtime/darwinia/src/pallets/system.rs
```diff
@@ -8,8 +8,26 @@ use crate::{weights::frame_system::WeightInfo, *};
...
+			// Ensure there is no new affirmation
+			// Then apply the migration (runtime upgrade) at UTC 07-30-2021, 10:00 AM
+			// For the started games, we expected the rounds could be finished during 05:00 AM ~ 10:00 AM
+			// TODO: remove this filter in the next runtime upgrade to restart the bridge
+			Call::EthereumRelay(darwinia_ethereum_relay::Call::affirm(..))
+				if System::block_number() > 4367266 =>
+			{
+				false
+			}
+			_ => true,
```
HackFisher, line 29: @WoeOm @sxlwar wormhole might need companion PR to disable this and tell the user with the infomation.

HackFisher: @AurevoirXavier I didn't get why we need to disable this, why can not just remove all existing affirmation when runtime upgrade

AurevoirXavier: Every `affirm` or `dispute_and_affirm` need to stake some fund, which uses for reward/slashing.
Once we remove the `affirmation` which will lock the fund permanently.

There is another way, remove all the storages (because they are related). Also, `take_storage_value(Stakers)` then do `Ring::remove_lock` for every staker.

HackFisher: > There is another way, remove all the storages (because they are related). Also, take_storage_value(Stakers) then do Ring::remove_lock for every staker.

Let's do it this way, this will eliminate extra workloads for wormhole & bridger & community support

HackFisher:
Related https://github.com/darwinia-network/darwinia-common/pull/721#issuecomment-887360337

HackFisher:
darwinia-common compare with last release:

https://github.com/darwinia-network/darwinia-common/compare/darwinia-v0.11.0...darwinia-v0.11.1-fork

HackFisher (approved):

PR #706, opened 2021-08-13
Drop `darwinia-cli`

AurevoirXavier:
1. Companion for https://github.com/darwinia-network/darwinia-common/pull/745
2. Remove `is_darwinia`. Because darwinia is the default option.


AurevoirXavier:
@AsceticBear @HackFisher, please review.

cli/src/command.rs
```diff
@@ -94,42 +92,31 @@ impl SubstrateCli for Cli {
...
-			)?),
-			path => Box::new(DarwiniaChainSpec::from_json_file(
-				std::path::PathBuf::from(path),
-			)?),
+			path => {
+				let path = PathBuf::from(path);
+				let chain_spec = Box::new(DarwiniaChainSpec::from_json_file(path.clone())?)
+					as Box<dyn ChainSpec>;
+
+				if self.run.force_crab || chain_spec.is_crab() {
```
HackFisher, line 100: Do we need `chain_spec.is_crab()` here? Is `crab-dev` spec `is_crab` true?

AurevoirXavier: Yes. CrabChainSpec shares with all Crab runtimes.

HackFisher: crab-dev should not use json file, right?

AurevoirXavier: Nope, maybe you want to make a testnet on different machines that you will need to export the chain state to a json file.

HackFisher (approved):

PR #721, opened 2021-08-26
Use Primitives Directly

AurevoirXavier:


AurevoirXavier:
> LGTM, but need some help explaining the storage prefix checking log.

These pallets using the same storage prefix, bot didn't catch this corner case.

PR #779, opened 2021-11-10
Improve CI

### LINKED ISSUE #766: Improve Release CI
- [ ] Change srtool use Subwasm or Srtool actions
- [ ] https://github.com/darwinia-network/darwinia/issues/771
- [ ] Split build rust docs to other actions
- [ ] https://github.com/darwinia-network/darwinia/issues/775
- [x] https://github.com/darwinia-network/darwinia/issues/708

fewensa:
Closes #766  ~~#777~~


AurevoirXavier:
#777 Require polkadot-v0.9.12. Keep using compact for current codebase.

fewensa:
https://github.com/fewensa/darwinia/actions/runs/1446989386

HackFisher (approved):
@AurevoirXavier

fewensa:
1. Store shared file, includes binary/wasm files, compiled by step [`build-package`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L9), use it in [`build-subwasm`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L77-L80) and [`build-image`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L117-L120)
2. Edit `.maintain/release-template.md` file. The template supports the use of variables, format: `{{ VAR_ANEM }}`, all available variables set by job [`Generate template data`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L217-L228) .

AurevoirXavier:
> 1. Store shared file, includes binary/wasm files, compiled by step [`build-package`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L9), use it in [`build-subwasm`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L77-L80) and [`build-image`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L117-L120)
> 2. Edit `.maintain/release-template.md` file. The template supports use variables, format: `{{ VAR_NAME }}`, all available variables set by job [`Generate template data`](https://github.com/darwinia-network/darwinia/blob/6099cf4d703b95343db148f3ce74c36c67ce4494/.github/workflows/release.yml#L217-L228) .

Did you clean up the artifact after using it?

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
 
 jobs:
-  build-binary:
-    name: Build binary
+
+  build-package:
+    name: Build package
     runs-on: ubuntu-latest
+    container:
+      image: ubuntu:16.04
```
wi1dcard, line 13: Why not specify `runs-on` to the Ubuntu version you want?

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
 
-  build-wasm:
-    name: Build wasm
+  build-subwasm:
+    name: Generate subwasm info
     runs-on: ubuntu-latest
+    container:
+      # Currently, Same with srtool version v0.9.18
+      # https://github.com/paritytech/srtool/blob/v0.9.18/Dockerfile#L1
+      image: ubuntu:20.04
```
wi1dcard, line 64: Same, `runs-on: ubuntu-latest` currently points to Ubuntu 20.04. Can remove this.

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
-      - name: Build darwinia
-        env:
-          RUST_TOOLCHAIN: nightly-2021-04-22
+      - name: Install dependencies
         run: |
-          .maintain/setup-darwinia.sh
+          apt update -y
+          apt install --no-install-recommends -y \
+            git curl ca-certificates \
+            make gcc g++ clang-4.0
```
wi1dcard, line 22: This step might be no longer necessary once removing the `container` section.

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
-      - uses: actions/checkout@v2
-      - name: Build wasm
-        env:
-          RUST_TOOLCHAIN: nightly-2021-04-22
+      - name: Install dependencies
         run: |
-          .maintain/setup-wasm.sh
+          apt update -y
+          apt install --no-install-recommends -y \
+            git curl ca-certificates
```
wi1dcard, line 74: Same, maybe unnecessary.

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
         with:
-          name: darwinia-artifact
-          path: deploy
+          length: 7
 
-      - name: Setup image
+      - name: Build and publish docker image
         env:
-          REGISTRY_HOST: quay.io
+          SHA: ${{ env.SHA }}
```
wi1dcard, line 137: Unnecessary?

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
         run: |
-          .maintain/setup-image.sh
+          chmod +x shared/darwinia
```
wi1dcard, line 139: The binary will automatically get the executive permission after `cargo build`. Is this required?

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
 
       - uses: actions/download-artifact@v2
         with:
           name: darwinia-artifact
-          path: deploy
+          path: shared
+
+      - name: Prepare deploy folder
+        run: |
+          chmod +x shared/darwinia
```
wi1dcard, line 171: Same

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          DRAFT_RELEASE: "false"
-          PRE_RELEASE: "false"
-          CHANGELOG_FILE: "CHANGELOG.md"
-          ALLOW_EMPTY_CHANGELOG: "false"
+          mv crab-runtime/*.compact.wasm ../deploy/
+          mv crab-runtime/*.compact.json ../deploy/
+          mv darwinia-runtime/*.compact.wasm ../deploy/
+          mv darwinia-runtime/*.compact.json ../deploy/
```
wi1dcard, line 181: I don't recommend using `cd` in CI scripts (especially having `../` together), as it might cause people confused about the context and working directory. For example:

- Other developers might not notice the `cd` above while adding more filesystem commands.
- It's hard to know "where I am" and "where is this file going" when I saw `../deploy/` only.

Instead, I personally would use something like (not tested):

```
mkdir deploy
tar cf darwinia.tar.zst -C shared darwinia -I pzstd
mv darwinia.tar.zst deploy/
mv shared/crab-runtime/*.compact.{wasm,json} deploy/
```

Also, there's a small sugar `*.compact.{wasm,json}`. I forget whether that's a zsh or bash feature. Confirm before you use this.

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
+#          mv crab-runtime/*.compact.compressed.wasm ../deploy/
+#          mv crab-runtime/*.compact.compressed.json ../deploy/
+#          mv darwinia-runtime/*.compact.compressed.wasm ../deploy/
+#          mv darwinia-runtime/*.compact.compressed.json ../deploy/
+
+      - name: Hash file
+        run: |
+          cd deploy
+          md5sum * > ../md5sums.txt
+          mv ../md5sums.txt ./
```
wi1dcard, line 193: Same too.

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
+            --arg darwinia_proposal_compact "$PROPOSAL_DARWINIA_COMPACT" \
+            --arg image_tag "$IMAGE_PUSH_NAME_TAG" \
+            --arg image_sha "$IMAGE_PUSH_NAME_SHA" \
+            '{
+              crab_proposal_compact: $crab_proposal_compact,
+              darwinia_proposal_compact: $darwinia_proposal_compact,
+              image_tag: $image_tag,
+              image_sha: $image_sha
+            }' )
+          echo "$JSON" > shared/template_data.json
```
wi1dcard, line 227: Can be simplified to `jq > shared/template_data.json`. Use `tee` if you want this to be outputted to CI logs (stdout).

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
+          length: 7
 
-      - uses: 8398a7/action-slack@v3
-        if: failure()
+      - name: Generate template data
+        run: |
+          PROPOSAL_CRAB_COMPACT=$(cat deploy/crab-runtime.compact.json | jq -r .proposal_hash)
+          PROPOSAL_DARWINIA_COMPACT=$(cat deploy/darwinia-runtime.compact.json | jq -r .proposal_hash)
+
+          # # not open for now. https://github.com/darwinia-network/darwinia/pull/779#issuecomment-964911852
```
wi1dcard, line 207: Two sharps.

wi1dcard: And the indents here don't consist with the other `not open for now.` comments. Please make sure to use a consistent indentation style.

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
         with:
-          status: ${{ job.status }}
-          fields: repo,author,eventName,workflow,ref,commit
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_INCOMING_WEBHOOK_URL }}
+          datafile: shared/template_data.json
+          path: .maintain/release-template.md
+
+      - name: Publish github release
+        uses: ncipollo/release-action@v1
```
wi1dcard, line 236: Any reason not using the more common one? https://github.com/softprops/action-gh-release

.github/workflows/release.yml
```diff
@@ -5,141 +5,246 @@ on:
...
+
+  clean-artifacts:
+    name: Clean artifacts
+    needs: [complete-build]
+    if: always()
+    steps:
+      - uses: geekyeggo/delete-artifact@v1
+        with:
+          name: darwinia-artifact
+    runs-on: ubuntu-latest
```
wi1dcard, line 250: With our flavor, `runs-on` should be on the top.

.github/workflows/rustdocs.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          toolchain: nightly-2021-04-22
+          target: wasm32-unknown-unknown
+          default: true
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+
+      - name: Install vercel
```
wi1dcard, line 26: Accurately this should be `vercel-cli`.

.github/workflows/rustdocs.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: npm i -g vercel
+
+      - name: Build rust docs
+        run: cargo doc --all --no-deps --release
+
+      - name: Deploy rust docs
+        run: |
+          mv vercel.json target/doc
+          mv target/doc target/${GITHUB_REPOSITORY#*/}
+          cd target/${GITHUB_REPOSITORY#*/}
```
wi1dcard, line 36: As above mentioned, a CLI argument or flag to set the directory name instead of moving files around would be great (if `vercel-cli` provide this).

Otherwise, better add a line of comment if `vercel-cli` requires a specific directory.

.maintain/README_zh_CN.md
```diff
@@ -30,193 +30,31 @@ cargo test
...
+Darwinia 直接采用 subwasm 来进行 runtime wasm proposal hash 计算, 采用与 srtool 相同的环境, 在计算完成后, 会输出变量到 release page 模板中, 视需要输出到页面, 同时也会将计算出的 json 存储为文件上传到 release page.
 
 ### 验证
 
-如果想要自行验证 wasm hash, 可以通过 srtool 来进行 darwinia/crab runtime.
-
-> 如上方阐述, 存在 [paritytech/srtool](https://hub.docker.com/r/paritytech/srtool) 以及 [chevdor/srtool](https://gitlab.com/chevdor/srtool) 两个库, Darwinia 使用的是 paritytech/srtool, 如果要自行验证也应该使用 paritytech/srtool, 保证自行编译使用的版本和 Darwinia 所使用的相同即可.
-
-为了能够正确编译, 你需要安装以下软件
+如果想要自行验证 wasm hash, 可以通过 polkadot 钱包在上传 wasm 后会得到一个 proposal hash, 然后与 CI 发布的进行对比.
```
wi1dcard, line 52: Better correcting all the capitalization in this doc, such as "Polkadot", "GitHub", etc.

.maintain/optimize-cache.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+#
+
+set -x
```
wi1dcard, line 4: `set -xe` might be a better choice (stops running the following commands if any error occurs).

.maintain/release-template.md
```diff
@@ -0,0 +1,10 @@
+Darwinia
+===
+
+This is a release template
```
wi1dcard, line 4: The description here needs to be updated.

wi1dcard (changes requested):

.github/workflows/release.yml
fewensa: There we need to keep this, because we want to compile the code in a unified and clean system

.github/workflows/release.yml
fewensa: There I can we can remove it.

.github/workflows/release.yml
fewensa: We need to keep using the docker container, so this is required.

.github/workflows/release.yml
fewensa: Right

.github/workflows/release.yml
fewensa: A question, When removed this.
https://github.com/darwinia-network/darwinia/blob/fd088d17f0451fddf020dee8affe9fa8ef3cefe0/.github/workflows/release.yml#L89
There is works?

.github/workflows/release.yml
fewensa: I'm not sure, This is the official suggestions, https://github.com/benjlevesque/short-sha#usage

.github/workflows/release.yml
fewensa: If run this command 
```
md5sum deploy/* > deploy/md5sum.txt
```
the output likes
```
12345    deploy/name.txt
```
It's ugly, So I think there can be change to

```
cd deploy
md5sum * > md5sums.txt
# or
md5sum * | tee md5sums.txt
```

.github/workflows/rustdocs.yml
fewensa: Yes, This step is required, I have already written a document in notion.

.maintain/optimize-cache.sh
fewensa: No, If it is the first time to execute, it is possible that the directory does not exist or there is no compiled package, so it is allowed to fail here

wi1dcard: Then you might want to add some bash parameter that makes the shell always return the exit code 0, or explicitly ignore the exit code from the caller side like `./optimize-cache.sh || exit 0`.

.github/workflows/release.yml
wi1dcard: You can use `(cd deploy && md5sum * > sums.txt)` and another problem is that md5 is now considered not secure anymore. `sha256 sum` would be a better alternative.

.github/workflows/release.yml
wi1dcard: Please confirm that, I don't think it is required setting the same env var with its own value.

.github/workflows/release.yml
wi1dcard: I can't think of any reason for now that this will stop working after we change to use the native `ubuntu-latest`.

.github/workflows/rustdocs.yml
wi1dcard: So I prefer adding a comment here. But that's just my style. It depends on you.

fewensa:
Example:
https://github.com/fewensa/darwinia/actions/runs/1456743431
https://github.com/fewensa/darwinia/releases/tag/v0.90.15

.maintain/release-template.md
fewensa: The real content needs to be modified before publishing

AurevoirXavier (approved):

PR #798, opened 2021-11-24
Use bz2 instead zstd

fewensa:


AurevoirXavier:
I mean, keep both versions.

But let me check the size. Maybe we can just post the entire binary there without compressing it.

AurevoirXavier:
Let's use binary.

AurevoirXavier:
Add bz2 and keep zst as well

cc @fewensa

HackFisher (approved):

PR #873, opened 2022-05-05
Remove `Claim` from Crab

### LINKED ISSUE #870: Remove claim pallet in Crab's next runtime
And the ui will be removed after that, all the unclaimed CRAB tokens will be claimed using new solution, aka. using airdrop tools.

And, the token balances of unclaimed token will need to be migrated to a multisig address for airdrop tool to use.

AurevoirXavier:
Close #870.

~~And there is a referendum on the way.~~
~~To decide how to deal with the un-claims. Burn or move to treasury account.~~
> https://crab.subscan.io/referenda/46


runtime/crab/src/migrations.rs
```diff
@@ -1,24 +1,48 @@
...
+	RuntimeBlockWeights::get().max_block
 }
 
 pub struct CustomOnRuntimeUpgrade;
 impl OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 	#[cfg(feature = "try-runtime")]
 	fn pre_upgrade() -> Result<(), &'static str> {
+		let claims_pallet_account = PalletId(*b"da/claim").into_account();
+
+		assert!(Ring::free_balance(&claims_pallet_account) != 0);
```
AsceticBear, line 35: Should be better to check the `claims_pallet_account` locked value here?

AurevoirXavier: `free_balance` means all except reserve.

AsceticBear: As my understanding, locked_balance is also part of the free_balance, we just check `!= 0`, what if `locked_balance` isn't `1 * COIN`.

AurevoirXavier: I have checked from chain state.

HackFisher (approved):
Conflicts exists.

HackFisher:
@AurevoirXavier Compile failed

PR #875, opened 2022-05-07
Shorten Feemarket Slot

AsceticBear:


AurevoirXavier:
Cherrypicked in #874.

PR #878, opened 2022-05-16
Crab <> Crab Parachain pallets

furoxr:
- [x]  bridge-parachain
- [x]  bridge-grandpa for Kusama
- [x]  bridge-message
- [x]  bridge-dispatch
- [x]  fee market
- [x] api

### Related
1. https://github.com/darwinia-network/darwinia-bridges-substrate/pull/76


HackFisher (approved):

runtime/crab/src/pallets/bridge_grandpa.rs
```diff
@@ -23,3 +23,10 @@ impl Config<WithDarwiniaGrandpa> for Runtime {
 	type MaxRequests = MaxRequests;
 	type WeightInfo = ();
 }
+
+impl Config<WithKusamaGrandpa> for Runtime {
+	type BridgedChain = bp_kusama::Kusama;
+	type HeadersToKeep = HeadersToKeep;
```
AurevoirXavier, line 29: Should use `bp_kusama::DAYS`

runtime/crab/src/pallets/fee_market.rs
```diff
@@ -27,8 +29,8 @@ where
 }
 
 frame_support::parameter_types! {
-	pub const FeeMarketPalletId: PalletId = PalletId(*b"da/feemk");
-	pub const FeeMarketLockId: LockIdentifier = *b"da/feelf";
+	pub const DarwiniaFeeMarketLockId: LockIdentifier = *b"da/feelf";
```
AurevoirXavier, line 32: If we change this, we need a migration for balance lock.

AsceticBear: The original `FeeMarketLockId` is not updated, just rename it here IMO. He removed the unused palllet_id.

PR #892, opened 2022-05-31
Try Update GITHUB_CACHE_VERSION

hackfisher:


AurevoirXavier:
We can not use stable toolchain. https://github.com/darwinia-network/darwinia-common/issues/896

PR #914, opened 2022-06-22
Companions of `darwinia-v0.12.3`

AsceticBear:
## TODO

- [x] Add DVM to the Darwinia network
- [x] Add necessary migration
- [x] Check ED value
- [x] Clean the node side
- [x] Run dvm basic ts tests
- [x] Test evm-tracing works well


runtime/darwinia/src/pallets/bridge_messages.rs
```diff
@@ -9,6 +9,7 @@ use pallet_bridge_messages::Config;
 use pallet_fee_market::s2s::{
 	FeeMarketMessageAcceptedHandler, FeeMarketMessageConfirmedHandler, FeeMarketPayment,
 };
+use sp_core::H160;
```
AurevoirXavier, line 12: substrate dep

node/service/src/service/dvm.rs
```diff
@@ -115,7 +115,8 @@ where
 					substrate_backend.clone(),
 					dvm_backend.clone(),
 					3,
-					0,
+					// TODO: improve this later
+					if network == "Crab" { 4969901 } else { 9100000 },
```
AurevoirXavier, line 119: Check when Darwinia will perform the runtime upgrade.

AsceticBear: Yeah, we need to tweak this value when the release date is confirmed.

AurevoirXavier (approved):

PR #923, opened 2022-07-19
Release `v0.12.3-1`

AurevoirXavier:
- client + runtime upgrade for Crab
- client upgrade for Darwinia

### DIFF
--- a/runtime/darwinia/src/migrations.rs
+++ b/runtime/darwinia/src/migrations.rs
@@ -23,13 +23,6 @@ impl OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 }
 
 fn migrate() -> Weight {
-	use fp_storage::{EthereumStorageSchema, PALLET_ETHEREUM_SCHEMA};
-
-	frame_support::storage::unhashed::put::<EthereumStorageSchema>(
-		&PALLET_ETHEREUM_SCHEMA,
-		&EthereumStorageSchema::V3,
-	);
-
-	// 0
-	RuntimeBlockWeights::get().max_block
+	0
+	// RuntimeBlockWeights::get().max_block
 }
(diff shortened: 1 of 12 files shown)


runtime/darwinia/src/migrations.rs
```diff
@@ -23,13 +23,6 @@ impl OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 }
 
 fn migrate() -> Weight {
-	use fp_storage::{EthereumStorageSchema, PALLET_ETHEREUM_SCHEMA};
```
AsceticBear, line 26: We shouldn't delete this migration. Darwinia runtime has not upgraded to `0.12.3` yet.

HackFisher: This is OK only if we use the 1230 spec version for runtime upgrade instead of 1231, please make that clear.

HackFisher (changes requested):
Please also bump darwinia's spec version

AurevoirXavier:
> Please also bump darwinia's spec version

I suggest keeping using [1230](https://github.com/darwinia-network/darwinia/releases/tag/v0.12.3) runtime for Darwinia.
Treat this as a client-side only upgrade for Darwinia.
Then we can skip the unnecessary runtime upgrade testing.

HackFisher:
> > Please also bump darwinia's spec version
> 
> I suggest keeping using [1230](https://github.com/darwinia-network/darwinia/releases/tag/v0.12.3) runtime for Darwinia. Treat this as a client-side only upgrade for Darwinia. Then we can skip the unnecessary runtime upgrade testing.

This breaks the convention of runtime version and node version, looks not good to me.

We can still use this client even if the native version and runtime version does not match anyway.

HackFisher (approved):

HackFisher:
And, The removal of migration actually affect darwinia runtime scope especially when some node run this patch version before runtime upgrade.

PR #969, opened 2023-03-08
Import from https://github.com/darwinia-network/darwinia-2.0

AurevoirXavier:
- https://github.com/darwinia-network/darwinia-2.0/issues/9

PR #978, opened 2023-03-08
Fix release CI

fewensa:


.github/workflows/staging.yml
```diff
@@ -31,6 +31,12 @@ jobs:
         with:
           toolchain: nightly
 
+      - name: Cache target
+        uses: actions/cache@v2
+        with:
+          path: target
```
AurevoirXavier, line 37: I recommend using a fresh build for release. Any cache might cause a problem.

Is this only for staging?

.github/release-template.md
```diff
@@ -38,5 +38,5 @@ docker pull ghcr.io/darwinia-network/darwinia-2.0:sha-{{ sha }}
 | Pangoro  | `{{ pangoro_proposal_compressed }}`  | `{{ pangoro_runtime_version }}`  |
 
 ## Changelog
-[Darwinia 2.0](https://github.com/darwinia-network/darwinia-2.0/issues/9)
+[Darwinia 2.0](https://github.com/darwinia-network/darwinia/issues/9)
```
boundless-forest, line 41: Please changes this link to https://github.com/darwinia-network/darwinia/pull/969

.github/workflows/staging.yml
fewensa: Okay, I agree

fewensa: What do you think about cache in release CI? If remove it from staging I think also remove it for release CI.

AurevoirXavier: You can keep it for staging.

But don't cache it for release.

.github/release-template.md
AurevoirXavier: Anyway, this will be updated in the final release PR.

hackfisher (approved):

AurevoirXavier:
Are we waiting for https://github.com/darwinia-network/darwinia/actions/runs/4361727015?

fewensa:
Not need

PR #983, opened 2023-03-08
Update migration state tests

boundless-forest:


AurevoirXavier (changes requested):
Figure out why the test failed.

And I think we can delete the Crab test and add the Darwinia test.

boundless-forest:
> Figure out why the test failed.

The export state wasn't updated in the CI, that's the reason why the test failed before.

> And I think we can delete the Crab test and add the Darwinia test.

Let's keep it for another week and see if there are any problems with state migration after the Crab2.0 launch. If all goes well, let's replace it after you export the darwinia state.

PR #989, opened 2023-03-08
Disable state check

AurevoirXavier:


boundless-forest:
Suggest keeping it until the darwinia 2.0 main network launch. We may tweak the state processor before we launch darwinia. This test will ensure that the core migration logic is correct.

PR #993, opened 2023-03-09
Optimize bls precompile by arkworks lib

hujw77:
### Test on local
fast_aggregate_verify: 990ms -> 560ms

### Test on pangoro
TBD


Cargo.toml
```diff
@@ -23,12 +23,20 @@ repository = "https://github.com/darwinia-network/darwinia"
 version    = "6.0.0"
 
 [workspace.dependencies]
+# arkworks
```
AurevoirXavier, line 26: These should not be workspace dependencies.

They weren't used widely.

precompile/ark-bls12-381/src/bls/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Darwinia. If not, see <https://www.gnu.org/licenses/>.
+//
+// Inspired from https://github.com/w3f/apk-proofs/blob/main/bw6/src/bls/mod.rs
+
+use ark_bls12_381::{
```
AurevoirXavier, line 21: Fmt

shell.nix
```diff
@@ -0,0 +1,26 @@
+let
+  mozillaOverlay =
+    import (builtins.fetchGit {
+      url = "https://github.com/mozilla/nixpkgs-mozilla.git";
+      rev = "78e723925daf5c9e8d0a1837ec27059e61649cb6";
+    });
+  nixpkgs = import <nixpkgs> { overlays = [ mozillaOverlay ]; };
+  rust-nightly = with nixpkgs; ((rustChannelOf { date = "2022-11-15"; channel = "nightly"; }).rust.override {
```
AurevoirXavier, line 8: We already have rust-toolchain.toml to control the version. Any specific version here will be overriden.

shell.nix
```diff
@@ -0,0 +1,26 @@
...
+  ];
+  buildInputs = [
+    clang
+    rocksdb
+  ];
+
+  RUST_SRC_PATH = "${rust-nightly}/lib/rustlib/src/rust/src";
+  LIBCLANG_PATH = "${llvmPackages.libclang.lib}/lib";
+  PROTOC = "${protobuf}/bin/protoc";
+  ROCKSDB_LIB_DIR = "${rocksdb}/lib";
```
AurevoirXavier, line 25: We'd also need `OPENSSL_LIB` if build for all-features.

AurevoirXavier:
But it looks like that 560ms is still unacceptable.

Any testing results?

https://github.com/darwinia-network/darwinia/blob/5ca9e8d4d23e2a6db28334f207c7e743df7bb002/runtime/pangoro/src/pallets/system.rs#L30-L31

shell.nix
hujw77: I use nixos for dev, just in the nix way.

AurevoirXavier: I mean, `cargo` will override this. Unless you use `rustc` to compile the project instead of `cargo`. Otherwise, this version is useless.

precompile/ark-bls12-381/src/bls/mod.rs
hujw77: `cargo fmt` makes no change.

hujw77:
> But it looks like that 560ms is still unacceptable.
> 
> Any testing results?
> 
> https://github.com/darwinia-network/darwinia/blob/5ca9e8d4d23e2a6db28334f207c7e743df7bb002/runtime/pangoro/src/pallets/system.rs#L30-L31

I want to test on Pangoro to see the result.
If it's still > 500ms, I will try to other optimize solution.

shell.nix
hujw77: Nixpkg is used for a package manager, there is no need for another package manager to deal with it.

runtime/darwinia/src/pallets/evm.rs
```diff
@@ -117,6 +119,8 @@ where
 				Some(<darwinia_precompile_deposit::Deposit<Runtime>>::execute(handle)),
 			a if a == addr(1537) =>
 				Some(<darwinia_precompile_staking::Staking<Runtime>>::execute(handle)),
+			a if a == addr(2047) =>
```
boundless-forest, line 122: Please change this to 2049, `[2048..) reserved for the experimental precompiles.`

Cargo.toml
```diff
@@ -60,6 +60,7 @@ darwinia-ecdsa-authority          = { default-features = false, path = "pallet/e
 darwinia-message-gadget           = { default-features = false, path = "pallet/message-gadget" }
 darwinia-message-transact         = { default-features = false, path = "pallet/message-transact" }
 darwinia-precompile-assets        = { default-features = false, path = "precompile/assets" }
+darwinia-precompile-ark-bls12-381 = { default-features = false, path = "precompile/ark-bls12-381" }
```
boundless-forest, line 63: Delete `darwinia-precompile-bls12-381` crate, which is not used anymore.

precompile/bls12-381/src/bls/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+		curve_maps::wb::WBMap, map_to_curve_hasher::MapToCurveBasedHasher, HashToCurve,
+		HashToCurveError,
+	},
+	models::short_weierstrass::Projective,
+	pairing::Pairing,
+	AffineRepr, CurveGroup,
+};
+use ark_ff::{field_hashers::DefaultFieldHasher, Zero};
+use ark_serialize::*;
+use ark_std::vec::Vec;
```
AurevoirXavier, line 38: Sometimes this is `ark_std::vec::Vec`. And sometimes this is `sp_std::vec::Vec`. Can we only use one here?

hujw77: Let me see the difference between them.

hujw77: They are same, use `sp_set` only

precompile/bls12-381/src/lib.rs
```diff
@@ -18,25 +18,42 @@
...
 // core
 use core::marker::PhantomData;
+// bls
+use bls::{hash_to_curve_g2, PublicKey, Signature};
 // moonbeam
 use precompile_utils::prelude::*;
 // substrate
 use sp_std::prelude::*;
 
+pub(crate) const BLS_ESTIMATED_COST: u64 = 100_000;
```
boundless-forest, line 32: Let's bump this value bigger, the currently EthBlockGasLimit is 10_000_000 gas. Let change this to the half of the MAX to avoid other big-gas transaction being package the same block.

boundless-forest: Let me fix it. Give me a sec.

hujw77: No, just keep it, If the pangoro test passed, I will benchmark the cost.

boundless-forest: Okay. Make sense to me.

boundless-forest (approved):

AurevoirXavier:
Don't merge; I need some review later.

hujw77:
Local 900ms -> 410 ms
CPU: 3.6GHz

PR #994, opened 2023-03-09
Loosen genesis reservation rule

AurevoirXavier:


boundless-forest (dismissed):

hackfisher (commented):
```
			if let Some(i) = <Identities<T>>::take(&from) {
				migration::put_storage_value(
					b"Identity",
					b"IdentityOf",
					&Twox64Concat::hash(&to.encode()),
					i,
				);
			}
```

Do we need to take care the actual deposit recording the reserve in pallets?
For example:

https://github.com/paritytech/substrate/blob/77c675b15cd805e5a9d61345763655ad8941591e/frame/identity/src/types.rs#L385-L386

AurevoirXavier:
> ```
> 			if let Some(i) = <Identities<T>>::take(&from) {
> 				migration::put_storage_value(
> 					b"Identity",
> 					b"IdentityOf",
> 					&Twox64Concat::hash(&to.encode()),
> 					i,
> 				);
> 			}
> ```
> 
> 
>      
>   
> 
> Do we need to take care the actual deposit recording the reserve in pallets?
> For example:
> 
> 
> https://github.com/paritytech/substrate/blob/77c675b15cd805e5a9d61345763655ad8941591e/frame/identity/src/types.rs#L385-L386

No need to.

Other wise we need to set all registrars' deposit to 0.

https://github.com/paritytech/substrate/blob/77c675b15cd805e5a9d61345763655ad8941591e/frame/identity/src/types.rs#L402

hackfisher:
Yes, that's specific to pallet logics who was using the reserve.

It is important to notice that, in production, err of 100% unreserve could happen in this PR's case, and is ignored without Err.

https://github.com/paritytech/substrate/blob/77c675b15cd805e5a9d61345763655ad8941591e/frame/identity/src/lib.rs#L360-L361

hackfisher:
Add one more comment on this, adoption of `reserve` is a bad design for substrate template pallets, cause it will cause extra burden for keep the ledger consistency between account data and specific pallets's data who use reserve, better way would be just transfer the reserve to pallet account, and let the pallet manage it by themself.

hackfisher (approved):
LGTM

PR #997, opened 2023-03-09
Release `v6.0.0-1`

aurexav:


runtime/darwinia/src/lib.rs
```diff
@@ -83,7 +83,7 @@ pub const VERSION: sp_version::RuntimeVersion = sp_version::RuntimeVersion {
 	spec_name: sp_runtime::create_runtime_str!("Darwinia2"),
 	impl_name: sp_runtime::create_runtime_str!("DarwiniaOfficialRust"),
 	authoring_version: 0,
-	spec_version: 6_0_0_0,
+	spec_version: 6_0_0_1,
```
boundless-forest, line 86: We do not need to bump darwinia version? as it has not been launched yet.

AurevoirXavier: Keep the same with the crate version.

hackfisher: How about pangolin?

AurevoirXavier: Pangolin is a testnet, I think I'll upgrade it frequently.

PR #1005, opened 2023-03-09
Fix docker image

fewensa:


.maintain/docker/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM debian:stable-slim
 
+RUN apt update && apt install -y --no-install-recommends ca-certificates && \
```
AurevoirXavier, line 3: I think we should use apt-get in docker env.

PR #1009, opened 2023-03-10
Optimize BLS precompile part 2

hujw77:
### Local
CPU: 3.6GHz
precompile fast_aggregate_verify: 410 ms -> 160ms

```diff
- deserialize_compressed
+ deserialize_compressed_unchecked
```

We validate the bls pubkeys in [solidity](https://github.com/darwinia-network/darwinia-messages-sol/blob/3732553161bc9692d6b8ce418316479289c97b23/contracts/bridge/src/truth/eth/BeaconLightClient.sol#LL154C17-L154C46).
So remove the bls pubkey validate check in precompile. refer: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.5


boundless-forest (approved):
Brilliant idea. Let's have a try!

PR #1010, opened 2023-03-10
Auto claim unstaking

AurevoirXavier:
Auto claim unstaking if it is expired while migrating.

### DIFF
--- a/pallet/account-migration/src/lib.rs
+++ b/pallet/account-migration/src/lib.rs
@@ -442,7 +442,7 @@ pub mod pallet {
 
 				migration::put_storage_value(b"Identity", b"Registrars", &[], rs);
 			}
-			if let Some(l) = <Ledgers<T>>::take(&from) {
+			if let Some(mut l) = <Ledgers<T>>::take(&from) {
 				if let Some(ds) = <Deposits<T>>::take(&from) {
 					<pallet_balances::Pallet<T> as Currency<_>>::transfer(
 						&to,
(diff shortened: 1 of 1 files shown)


pallet/account-migration/src/lib.rs
```diff
@@ -457,14 +457,41 @@ pub mod pallet {
...
 				)?;
 
-				let sum = l.staked_kton + l.unstaking_kton.iter().map(|(k, _)| k).sum::<Balance>();
+				let sum = l.staked_kton
+					+ l.unstaking_kton
+						.iter()
+						.filter_map(
+							|(amount, expired_time)| {
+								if expired_time <= &now {
+									None
```
hackfisher, line 487: Does the ledger here get updated?

AurevoirXavier: I just wanted to update this. Wait a min.

hackfisher: Another thing I would like to check is that, for the migrations before this upgrate, the unstakings that moved to staking pot and stayed in ledger, can they claim their unstaking?

Do we need to fix for them?

AurevoirXavier: Yes. They can claim them immediately. @xiaoch05 did this yesterday.

AurevoirXavier:
@boundless-forest please also help review this.

pallet/account-migration/src/lib.rs
```diff
@@ -456,7 +456,13 @@ pub mod pallet {
 					);
 				}
 
+				let now = <frame_system::Pallet<T>>::block_number();
+
+				l.unstaking_ring.retain(|(_, t)| t > &now);
```
hackfisher, line 461: wait, where do the removed unstaking_ring go? not in free right?

hackfisher: OK, so free includes staked and unstaking, deposits.

boundless-forest: The unstaking part belong to the user now. It'll transfer to the `staking_pot` account later.

boundless-forest (approved):

PR #1022, opened 2023-03-15
Fix ecdsa-authority test

hujw77:


pallet/ecdsa-authority/src/primitives.rs
```diff
@@ -103,13 +103,22 @@ pub struct Commitment {
 
 #[test]
 fn eth_signable_message() {
+	fn domain_separator(chain_id: u64, spec_name: &[u8]) -> Hash {
```
AurevoirXavier, line 106: Maybe we can change chain id to u32 this time.

hujw77: Make sense

PR #1023, opened 2023-03-15
Fix vesting

AurevoirXavier:


runtime/crab/src/migration.rs
```diff
@@ -41,7 +75,48 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
...
+		// If `is_some` which means it hasn't been migrated yet.
+		// But actually, without this correction, the account migration will fail.
+		if unhashed::get::<VestingInfo>(&k).is_some() {
+			let v = vec![VestingInfo {
+				locked: (100_000_000_000_u128
+					- (CRAB_LAST_FINALIZE_HEIGHT - 8_421_033) as Balance * 1_u128)
+					* DECIMAL_OFFSET,
+				// Crab2's block time is twice longer than Crab.
+				per_block: 1 * DECIMAL_OFFSET * 2,
+				starting_block: 0,
```
boundless-forest, line 103: As the crab 2.0 has launched, so I think the `starting_block` should be the migration execute number rather than 0.

tool/state-processor/src/adjust.rs
```diff
@@ -35,11 +35,21 @@ impl Adjust for BalanceLock {
...
+		} else {
+			self.locked -= released;
+		}
+
 		self.locked.adjust();
-		self.per_block *= 2;
 		self.per_block.adjust();
-		self.starting_block =
-			self.starting_block.checked_sub(*NOW.read().unwrap()).unwrap_or_default();
+		self.per_block *= 2;
```
boundless-forest, line 51: The vesting module unlocked balance as the block number increase. 
This process is linked to the block number rather than timestamp. So I think we don't need to double this value.

runtime/crab/src/migration.rs
AurevoirXavier: Nope. I use the genesis state.

tool/state-processor/src/adjust.rs
AurevoirXavier: I don't think so.

For example, our team's vesting plan is to unlock all team reservations in 2024. Without this modification, the time will be 2026.

runtime/crab/src/migration.rs
```diff
@@ -41,7 +75,48 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
...
+	if let Ok(k) = array_bytes::hex2bytes(BROKEN_STORAGES[1]) {
+		// If `is_some` which means it hasn't been migrated yet.
+		// But actually, without this correction, the account migration will fail.
+		if unhashed::get::<VestingInfo>(&k).is_some() {
+			let v = vec![VestingInfo {
+				locked: (100_000_000_000_u128
+					- (CRAB_LAST_FINALIZE_HEIGHT - 8_421_033) as Balance * 1_u128)
+					* DECIMAL_OFFSET,
+				// Crab2's block time is twice longer than Crab.
+				per_block: 1 * DECIMAL_OFFSET * 2,
```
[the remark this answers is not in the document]
AurevoirXavier: Same.

runtime/crab/src/migration.rs
boundless-forest: Okay. I got it. But in this way, you need to update the `locked` field?  take the crab 2.0 current block number into account.

```
100_000_000_000_u128 - (CRAB_LAST_FINALIZE_HEIGHT - 8_421_033) * 1 - crab_2.0 number * 2
```

AurevoirXavier: Nope. Genesis state doesn't care about the future block number.

tool/state-processor/src/adjust.rs
boundless-forest: Thanks for your reply. This reason make sense to me.

hackfisher (approved):

boundless-forest (approved):

Issue #1040, opened 2023-02-08
Sorting out special accounts


boundless-forest:
```
migrate special account `SiblId(2023)` from `0x7369626ce7070000000000000000000000000000000000000000000000000000` to `0x7369626ce7070000000000000000000000000000`
migrate special account `PalletId(da/paais)` from `0x6d6f646c64612f70616169730000000000000000000000000000000000000000` to `0x6d6f646c64612f70616169730000000000000000`
migrate special account `SiblId(2004)` from `0x7369626cd4070000000000000000000000000000000000000000000000000000` to `0x7369626cd4070000000000000000000000000000`
migrate special account `PalletId(PotStake)` from `0x6d6f646c506f745374616b650000000000000000000000000000000000000000` to `0x6d6f646c506f745374616b650000000000000000`
migrate special account `PalletId(da/ethrl)` from `0x6d6f646c64612f657468726c0000000000000000000000000000000000000000` to `0x6d6f646c64612f657468726c0000000000000000`
migrate special account `PalletId(da/trsry)` from `0x6d6f646c64612f74727372790000000000000000000000000000000000000000` to `0x6d6f646c64612f74727372790000000000000000`
migrate special account `PalletId(da/crais)` from `0x6d6f646c64612f63726169730000000000000000000000000000000000000000` to `0x6d6f646c64612f63726169730000000000000000`
migrate special account `0x0000000000000000000000000000000000000000000000000000000000000000`
migrate special account `PalletId(da/ethfe)` from `0x6d6f646c64612f65746866650000000000000000000000000000000000000000` to `0x6d6f646c64612f65746866650000000000000000`
migrate special account `PalletId(py/trsry)` from `0x6d6f646c70792f74727372790000000000000000000000000000000000000000` to `0x6d6f646c70792f74727372790000000000000000`
migrate special account `PalletId(da/staki)` from `0x6d6f646c64612f7374616b690000000000000000000000000000000000000000` to `0x6d6f646c64612f7374616b690000000000000000`
migrate special account `PalletId(da/s2sba)` from `0x6d6f646c64612f73327362610000000000000000000000000000000000000000` to `0x6d6f646c64612f73327362610000000000000000`
migrate special account `SiblId(2000)` from `0x7369626cd0070000000000000000000000000000000000000000000000000000` to `0x7369626cd0070000000000000000000000000000`
[...]

PR #1047, opened 2023-03-20
More detail log

AurevoirXavier:
To make #1040 easier than before.


tool/state-processor/src/system/mod.rs
```diff
@@ -167,7 +168,7 @@ where
 					&blake2_128_concat_to_string(k),
 					a,
 				);
-			} else if let Some(k) = try_get_evm_address(&key) {
+			} else if let Some(k) = try_get_evm_address(&k) {
```
boundless-forest, line 171:
```suggestion
			} else if let Some(k) = try_get_evm_address(&k_64) {
```

PR #1050, opened 2023-03-22
Update testnets gas price

boundless-forest:


runtime/common/src/lib.rs
```diff
@@ -68,9 +68,7 @@ impl WeightToFeePolynomial for WeightToFee {
 	type Balance = Balance;
 
 	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-		// in Rococo, extrinsic base weight (smallest non-zero weight) is mapped to 1 MILLIUNIT:
-		// here, we map to 1/10 of that, or 1/10 MILLIUNIT
-		let p = MILLIUNIT / 10;
+		let p = MILLIUNIT;
```
AurevoirXavier, line 71: Any doc, why set it to milliunit?

runtime/pangoro/src/pallets/evm.rs
```diff
@@ -18,10 +18,13 @@
 
 // darwinia
 use crate::*;
+// substrate
+use frame_support::weights::WeightToFee;
```
AurevoirXavier, line 22: Can we put these inside the function body?

runtime/common/src/lib.rs
boundless-forest: According to the formula `GasPrice = NextFeeMultiplier * WeightToFee(WeightPerGas)`, that we use `SlowAdjustingFeeUpdate` as our main `FeeMultiplierUpdate` and I don't want to update it. `WeightPerGas` is a fixed value once the `BLOCKGASLIMIT` and `BLOCKWEIGHTLIMIT` are set. So the only way is to tweak the `WeightToFee ` function to adjust the parameters to get the gas price to fit our needs. The gas price calculated without touching the WeightToFee is `187804800`, which is too low in our case. that why I update this value.

PR #1078, opened 2023-03-29
Add more events and prevent duplicative submission

AurevoirXavier:


hackfisher (approved):

pallet/account-migration/src/lib.rs
```diff
@@ -129,7 +129,9 @@ pub mod pallet {
 	pub enum Event {
 		/// An account has been migrated.
 		Migrated { from: AccountId32, to: AccountId20 },
-		/// An multisig account has been migrated.
+		/// A new multisig account params was noted/recorded on-chain.
+		NewMultisigParamsNoted { from: AccountId32, to: MultisigParams },
```
boundless-forest, line 133: Suggest renaming `to: MultisigParams` to `param: MultisigParams`. The struct contain more than just `to` information.

pallet/account-migration/src/lib.rs
```diff
@@ -248,9 +255,13 @@ pub mod pallet {
...
 
-				Self::deposit_event(Event::MultisigMigrated { from, detail });
+				Self::deposit_event(Event::MultisigMigrated { from: from.clone(), detail });
 			} else {
-				<Multisigs<T>>::insert(from, detail);
+				<Multisigs<T>>::insert(&from, detail);
+			}
+
+			if let Some(to) = new_multisig_params {
+				Self::deposit_event(Event::NewMultisigParamsNoted { from, to });
```
boundless-forest, line 264: Should we deposit event in the case of `new_multisig_params` is `NONE`? Maybe we should.

AurevoirXavier: Nope

pallet/account-migration/src/lib.rs
AurevoirXavier: `param` looks confusing to me. The param is all about the `to`, the new multisig.

boundless-forest (approved):

PR #1093, opened 2023-04-06
Use default token conversion rate

boundless-forest:
Default value is enough with Fee market.


AurevoirXavier:
> Default value is enough with Fee market.

Can you elaborate on it more?

boundless-forest:
> > Default value is enough with Fee market.
> 
> Can you elaborate on it more?

This parameter is used to calculate the cross-chain fee in the original `pallet-bridger-message` design. In darwinia, we replace the fee calculation section with the fee market, which does not rely on the token conversion parameter.

PR #1098, opened 2023-04-12
Add `pallet-evm` origin tests

boundless-forest:
- https://github.com/darwinia-network/darwinia/blob/main/runtime/darwinia/src/pallets/evm.rs#L136

It's important that the pallet-evm call are not avaiable to the ordinary users. Add tests to make sure this restriction is not broken in the long development.


AurevoirXavier (dismissed):

boundless-forest:
> Can you add the origin link to the code?

Done

PR #1100, opened 2023-04-13
Remove unnecessary variable

AurevoirXavier:
https://github.com/chevdor/subwasm/pull/64


AurevoirXavier:
Need to bump the `SUBWASM_VERSION`.

PR #1113, opened 2023-04-20
Enable BLS precompile

boundless-forest:


hackfisher (approved):

node/src/chain_spec/darwinia.rs
```diff
@@ -35,7 +35,7 @@ use fp_evm::GenesisAccount;
 use sc_chain_spec::Properties;
 use sc_service::ChainType;
 use sc_telemetry::TelemetryEndpoints;
-use sp_core::H160;
+use sp_core::{crypto::UncheckedInto, H160};
```
AurevoirXavier, line 38: What happened?

boundless-forest: image

PR #1117, opened 2023-04-23
Add tracing apis

boundless-forest:
Add EVM tracing APIs to all darwinia's runtime.

TODO: Add the tracing test to ci @AurevoirXavier 

```sh
# setup node
$ cargo build --release -p darwinia --features pangolin-native,evm-tracing
$ ./target/release/darwinia --chain pangolin-dev --tmp --alice --tracing-api debug,trace

# run test
$ cd tests/ethereum
$ npm run test
```


AurevoirXavier (changes requested):
I don't like this style.

Can you make:

```
crab-evm-tracing = [
  "crab-native",
  ..
]
```

---

By the way, I believe that implementing an `evm-tracing` feature is feasible. This feature can be developed independently of the native components.

boundless-forest:
> By the way, I believe that implementing an evm-tracing feature is feasible. This feature can be developed independently of the native components.

Sound like the `try-runtime` or `runtime-benchmarks`?

AurevoirXavier:
> > By the way, I believe that implementing an evm-tracing feature is feasible. This feature can be developed independently of the native components.
> 
> Sound like the `try-runtime` or `runtime-benchmarks`?

Yep.

node/src/service/executors.rs
```diff
@@ -26,7 +26,9 @@ pub struct DarwiniaRuntimeExecutor;
 impl NativeExecutionDispatch for DarwiniaRuntimeExecutor {
 	#[cfg(feature = "runtime-benchmarks")]
 	type ExtendHostFunctions = frame_benchmarking::benchmarking::HostFunctions;
-	#[cfg(not(feature = "runtime-benchmarks"))]
+	#[cfg(feature = "evm-tracing")]
```
AurevoirXavier, line 29: What about you enabling runtime benchmark and evm tracing at the same time?

AurevoirXavier: But I think no one would do that.

boundless-forest: Exactly. evm-tracing is nothing more other than enabling a trace runtime API implementation. And I don't want to add this host function for a normal collator node.

node/src/frontier_service.rs
```diff
@@ -90,8 +104,70 @@ pub fn spawn_frontier_tasks<B, BE, C>(
...
+		// `debug` task if enabled. Essential.
+		// Proxies rpc requests to it's handler.
+		if let Some(debug_task) = debug_task {
+			task_manager.spawn_essential_handle().spawn(
+				"tracing_api-debug",
+				Some("eth-tracing"),
+				debug_task,
+			);
+		}
+		RpcRequesters { debug: debug_requester, trace: trace_filter_requester }
```
AurevoirXavier, line 167: We can enable the debug and trace at the same time?
What differences between them?

boundless-forest: > We can enable the debug and trace at the same time?

Yes, we can. the debug and trace are two types of rpc, enable `debug_transaction`, `debug_blockByNumber` and `trace_filter` respectively. Check out the ts test in this pr.

AurevoirXavier:
It looks like this break the origin evm test?

boundless-forest:
> It looks like this breaks the origin evm test?

The new evm tests(with tracing ts test) are a superset of the original evm test. This means that we don't need to run the original test separately. Setup the node with trace options, then run the new evm tests. If all the tests pass, this means that the evm works well.

PR #1125, opened 2023-04-26
Release `v6.2.0`

AurevoirXavier:


runtime/pangolin/src/pallets/ecdsa_authority.rs
```diff
@@ -29,7 +29,7 @@ static_assertions::const_assert!(MAX_PENDING_PERIOD > SYNC_INTERVAL);
 
 impl darwinia_ecdsa_authority::Config for Runtime {
 	type ChainId = <Self as pallet_evm::Config>::ChainId;
-	type MaxAuthorities = ConstU32<3>;
+	type MaxAuthorities = ConstU32<7>;
```
boundless-forest, line 32: @echo Please help check this value.

hackfisher: This max, can be 5?

AurevoirXavier: Yes, but it's just an upper limit. Every authority needs to be added/reviewed by the governance.

PR #1134, opened 2023-04-28
Fix  `try-runtime` on the darwinia network

boundless-forest:


AurevoirXavier:
Actually, this isn't a fix. It's just we haven't enabled the migration for Darwinia yet.

PR #1138, opened 2023-05-08
All fees go to treasury

AurevoirXavier:


runtime/common/src/lib.rs
```diff
@@ -97,13 +97,10 @@ where
...
 			use frame_support::traits::Imbalance;
 
-			// for fees, 80% are burned, 20% to the treasury
-			let (_, to_treasury) = fees.ration(80, 20);
-
 			// Balances pallet automatically burns dropped Negative Imbalances by decreasing
 			// total_supply accordingly
 			<pallet_treasury::Pallet<R> as frame_support::traits::OnUnbalanced<_>>::on_unbalanced(
-				to_treasury,
+				fees,
```
boundless-forest, line 103: Is it possible to add some runtime tests to this?

AurevoirXavier: I haven't tried that before.

PR #1139, opened 2023-05-09
Use workspace features

boundless-forest:
See [Inheriting a dependency from a workspace](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace)


AurevoirXavier:
I removed this in the previous PR.

I think the crate itself should manage the detail features.

boundless-forest:
> I suggest that we only manage the common dependency version in the workspace toml.

AurevoirXavier (changes requested):
Please add `features` after the `default-features = false`

PR #1146, opened 2023-05-16
Remove state processor things

AurevoirXavier:


runtime/pangoro/tests/tests.rs
```diff
@@ -18,10 +18,6 @@
 
 pub mod mock;
 
-darwinia_common_runtime::impl_account_migration_tests! {}
```
boundless-forest, line 21: Why remove this? This test is independent with state-processor.

AurevoirXavier: Since the state processor gets deleted, I think we won't make any changes to the state.

boundless-forest: This test has nothing to do with the state processor, the aim of this test is to ensure the correctness of the migration pallet. We may find there are some potential bugs in the migration pallet or be broken by the substrate upgrade in the future, this is possible.

boundless-forest (approved):

PR #1150, opened 2023-05-19
Update testnets bootnodes

fewensa:
testnets migrated to new server


AurevoirXavier:
We should update the json as well.

PR #1151, opened 2023-05-22
Deploy testnets tracing node

fewensa:


.github/workflows/release.yml
```diff
@@ -138,7 +138,7 @@ jobs:
   build-docker-image:
     name: Build Docker image
     runs-on: ubuntu-latest
-    needs: [build-package]
+    needs: [ build-package ]
```
AurevoirXavier, line 141: Suggest keeping the origin style

.github/workflows/release.yml
```diff
@@ -256,6 +256,29 @@ jobs:
...
+        if: startsWith(github.ref, 'refs/tags/pango')
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TKN_DARWINIA }}
+        run: |
+          gh workflow run \
+            trigger-tracing-node.yml \
+            --ref=main \
+            --repo=darwinia-network/devops \
+            --raw-field=type=testnets \
+            --raw-field="version=$GIT_TAG_NAME" \
```
AurevoirXavier, line 280: I think we don't need the `\` in the last line.

PR #1165, opened 2023-05-30
Add dispatch filter

boundless-forest:


runtime/crab/tests/mock.rs
```diff
@@ -55,6 +56,9 @@ impl ExtBuilder {
 		}
 		.assimilate_storage(&mut t)
 		.unwrap();
+		pallet_sudo::GenesisConfig::<Runtime> { key: Some(H160::from_low_u64_be(999).into()) }
```
AurevoirXavier, line 59: Why not use `ROOT` here?

boundless-forest: Nice suggestion! Let me fix it.

PR #1179, opened 2023-06-16
Update the weight `proof_size`

boundless-forest:


AurevoirXavier:
Is pangolin<>pangoro bridge down? @fewensa

PR #1193, opened 2023-06-28
Improve commands and use Substrate weights

AurevoirXavier:
- Improve command part
- Use Substrate weights instead of a copy from Polkadot
- Remove old migration


node/src/command.rs
```diff
@@ -438,7 +443,9 @@ pub fn run() -> Result<()> {
 		},
 		Some(Subcommand::PurgeChain(cmd)) => {
 			let runner = cli.create_runner(cmd)?;
+			let chain_spec = &runner.config().chain_spec;
 
+			set_default_ss58_version(chain_spec);
```
boundless-forest, line 448: Why need to `set_default_ss58_version` when purge chain?

node/src/command.rs
```diff
@@ -544,35 +559,39 @@ pub fn run() -> Result<()> {
 			use frame_benchmarking_cli::{BenchmarkCmd, SUBSTRATE_REFERENCE_HARDWARE};
 
 			let runner = cli.create_runner(&**cmd)?;
-			// Switch on the concrete benchmark sub-command-
+
```
boundless-forest, line 562: Suggest removing this empty line.

node/src/command.rs
AurevoirXavier: No specific reason. Just keep it the same with others. :)

PR #1204, opened 2023-07-12
Optimize CI

### LINKED ISSUE #1201: Rust cache
https://github.com/Swatinem/rust-cache

AurevoirXavier:
Optimize the compilation based on my latest knowledge.

Closes #1201.


.github/workflows/checks.yml
```diff
@@ -149,7 +158,7 @@ jobs:
           chmod u+x darwinia
           sudo mv darwinia /usr/bin
       - name: Launch darwinia
-        run: darwinia --chain pangolin-dev --tmp --alice --tracing-api debug,trace &
+        run: darwinia --chain pangolin-dev --tmp --alice --tracing-api debug,trace --frontier-backend-type sql &
```
boundless-forest, line 161: It's better to keep the original task without `--frontier-backend-type sql`, because not all of the collator nodes will upgrade their fronter db in the short term.

AurevoirXavier: IMO, only RPC node will use this. Collator doesn't care about this.

PR #1205, opened 2023-07-14
Frontier upgrade patch

boundless-forest:


pallet/message-transact/src/lib.rs
```diff
@@ -148,6 +148,16 @@ pub mod pallet {
 			};
 
 			let transaction_data: TransactionData = (&*transaction).into();
+			let (weight_limit, proof_size_base_cost) =
```
AurevoirXavier, line 151: It seems that this implementation has been copied from the upstream source.

node/src/frontier_service.rs
```diff
@@ -203,3 +207,48 @@ where
...
 }
+
+/// Create a Frontier backend.
+pub(crate) fn frontier_backend<B, BE, C>(
+	client: Arc<C>,
+	config: &sc_service::Configuration,
+	eth_rpc_config: EthRpcConfig,
+) -> Result<fc_db::Backend<B>, String>
+where
+	C: 'static
```
AurevoirXavier, line 218: Suggest following the order of function declaration: `B, BE, C`.

pallet/message-transact/src/lib.rs
AurevoirXavier: It would be better to link the source code here.

boundless-forest: Yes. I'll dig into deep to see if there is a way to reuse upstream function. Keep it now.

boundless-forest: The current way is fine with me. This piece of code is not like const pallet name, etc. which are tightly bound to something.

AurevoirXavier: Otherwise, I need an issue to keep track of this and we can update this in the future. Currently, my memory is not very reliable.

AurevoirXavier: #1206

Issue #1123, opened 2023-04-25
Disable Vesting calls, and remove this pallet after all current vesting accounts ends.


hackfisher:

AurevoirXavier:
Remove vesting for testnets first.

AurevoirXavier:
Can we move the vesting to EVM contract?

hackfisher:
> Can we move the vesting to EVM contract?

Yes, users/apps can deploy vesting smart contracts if they desire

PR #1214, opened 2023-07-20
Remove vesting from testnets

AurevoirXavier:
Part of #1123.


boundless-forest:
Darwinia:

image

Crab:

image


There are no storage items in the `Vestings` of the AccountMigration. So I think we can delete the vesting pallet for all runtimes at once, including darwina and crab. By the way, delete the migration logic and the migration integrate test.

AurevoirXavier:
Yesterday, there is still one item in `AccountMigration::Vesting`. It's fine to delete the `Vesting` now.

boundless-forest (approved):

PR #1235, opened 2023-08-07
Record `proof_size` cost for precompiles

### LINKED ISSUE #1233: Record external cost for the substrate pallet based precompile
The substrate pallet based precompile introduces extra proof_size cost, this part should also be charged under the new transaction fee machenisim.

boundless-forest:
Close #1233

### DIFF
--- a/precompile/assets/src/lib.rs
+++ b/precompile/assets/src/lib.rs
@@ -83,6 +83,7 @@ where
 	#[precompile::public("totalSupply()")]
 	#[precompile::view]
 	fn total_supply(handle: &mut impl PrecompileHandle) -> EvmResult<U256> {
+		// Record proof_size cost for total_issuance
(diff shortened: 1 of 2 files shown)


AurevoirXavier:
Is this only necessary for the asset pallet?

boundless-forest:
> precompile::view

And with DB reads or writes

AurevoirXavier:
> > precompile::view
> 
> And with DB reads or writes

Why https://github.com/darwinia-network/darwinia/blob/bb0bffca7f2a5d643e49a64b62f8e482b90d6c81/precompile/bls12-381/src/lib.rs#L59 use `record_cost`?

boundless-forest:
> use `record_cost`?

Nice catch. That's because this is a pure compution precompile, no db operation involved, so no extra proof_size cost there.

PR #1249, opened 2023-08-31
Release `v6.4.0`

AurevoirXavier:


runtime/crab/src/migration.rs
```diff
@@ -41,7 +41,16 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 }
 
 fn migrate() -> frame_support::weights::Weight {
-	frame_support::weights::Weight::zero()
+	// substrate
+	use frame_support::traits::StorageVersion;
```
boundless-forest, line 45: These changes are also needed for the darwinia runtime?

AurevoirXavier: No

runtime/crab/src/migration.rs
```diff
@@ -41,7 +41,16 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
...
+
+	// Some pallets ere added on chain after the migration.
+	// Thus, they never required the migration and we just missed to set the correct
+	// `StorageVersion`.
+	StorageVersion::new(4).put::<Scheduler>();
+	StorageVersion::new(1).put::<PolkadotXcm>();
```
jiguantong, line 51: What's the meaning of this line? What would happen if we don't set it?

AurevoirXavier: https://github.com/itering/internal-security-audit/pull/53

runtime/crab/src/migration.rs
boundless-forest:
```sh
2023-08-30 13:39:11.238  INFO main runtime::frame-support: ⚠️ XcmpQueue declares internal migrations (which *might* execute). On-chain `StorageVersion(3)` vs current storage version `StorageVersion(2)`
2023-08-30 13:39:11.238  INFO main runtime::frame-support: ⚠️ DmpQueue declares internal migrations (which *might* execute). On-chain `StorageVersion(2)` vs current storage version `StorageVersion(1)`
```
I analysed the darwinia try-runtime result and found that these two pallet also break the try-runtime process. If we do nothing at all, the darwinia runtime try-runtime still panic if I understand correctly.

AurevoirXavier: Please check the audit report.

boundless-forest (approved):

jiguantong (dismissed):

PR #1271, opened 2023-09-18
Prevent potential overflow

AurevoirXavier:
This PR only focuses on enhancing the robustness of the core/inflation libray, preventing any potential overflow.
It really doesn't care about the inputs.

All unit tests have been passed.

### DIFF
--- a/core/inflation/src/lib.rs
+++ b/core/inflation/src/lib.rs
@@ -45,8 +45,9 @@ pub const MILLISECS_PER_YEAR: Balance = (366 * 24 * 60 * 60) * 1000;
 ///
 /// Use `U94F34` here, because `2^94 > TOTAL_SUPPLY * 10^9`.
 pub fn in_period(unminted: Balance, period: Moment, elapsed: Moment) -> Option<Balance> {
-	let unminted_per_millisecs = U94F34::from_num(unminted) / MILLISECS_PER_YEAR;
-	let x = (unminted_per_millisecs * period).floor().to_num();
+	let unminted_per_millisecs = U94F34::checked_from_num(unminted)? / MILLISECS_PER_YEAR;
+	let x =
+		(unminted_per_millisecs.checked_mul(U94F34::checked_from_num(period)?)?).floor().to_num();
 	let years = (elapsed / MILLISECS_PER_YEAR + 1) as _;
 
 	inflate(x, years)
@@ -65,7 +66,7 @@ fn inflate(x: Balance, years: u8) -> Option<Balance> {
 	let ninety_nine = I95F33::from_num(99_u8) / 100_i128;
 	let pow = transcendental::pow::<I95F33, I95F33>(ninety_nine, sqrt).ok()?;
 	let ratio = I95F33::from_num(1_u8) - pow;
-	let inflation = I95F33::from_num(x) * ratio;
+	let inflation = I95F33::checked_from_num(x)? * ratio;
 
 	Some(inflation.floor().to_num())
 }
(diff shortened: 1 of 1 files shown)


core/inflation/src/lib.rs
```diff
@@ -75,6 +76,20 @@ fn inflate(x: Balance, years: u8) -> Option<Balance> {
...
 pub fn deposit_interest(amount: Balance, months: u8) -> Balance {
+	// The result of `((quot - 1) * precision + rem * precision / d)` is `197` when months is
+	// `12`.
+	//
+	// The default interest is `1_000`.
+	// So, we directly use `1_970_000` here instead `interest * 197 * 10^7`.
+	fn f(amount: U256, precision: U256, quot: U256, rem: U256, d: U256) -> Option<Balance> {
+		Some(
+			(amount.checked_mul(
+				precision.checked_mul(quot.checked_sub(1_u8.into())?)? + precision * rem / d,
```
boundless-forest, line 87: `precision * rem` : precision is a fixed U256 value and rem is a U256 value, comes from n and d. Does it necessary to use checked_mul here?

core/inflation/src/lib.rs
```diff
@@ -75,6 +76,20 @@ fn inflate(x: Balance, years: u8) -> Option<Balance> {
 /// Reference(s):
 /// - <https://github.com/evolutionlandorg/bank/blob/master/contracts/GringottsBank.sol#L280>
 pub fn deposit_interest(amount: Balance, months: u8) -> Balance {
+	// The result of `((quot - 1) * precision + rem * precision / d)` is `197` when months is
+	// `12`.
+	//
+	// The default interest is `1_000`.
+	// So, we directly use `1_970_000` here instead `interest * 197 * 10^7`.
```
boundless-forest, line 83: Very confused about this comment. How the `1_970_000` and `* 10^7` calculated here? If we are do multiple operation with 197 and 1_000, the result should be 197_000 instead of 1_970_000. 

Both he explanation here and in the evolutionland repo is not clear enough.

boundless-forest: @hackfisher  Please help review this.

AurevoirXavier: Without the knowledge from https://github.com/evolutionlandorg/bank/blob/master/contracts/GringottsBank.sol#L280, you can't understand it.

core/inflation/src/lib.rs
AurevoirXavier: `quot.checked_sub(1_u8.into())` is a var.

core/inflation/src/lib.rs
```diff
@@ -65,7 +66,7 @@ fn inflate(x: Balance, years: u8) -> Option<Balance> {
 	let ninety_nine = I95F33::from_num(99_u8) / 100_i128;
 	let pow = transcendental::pow::<I95F33, I95F33>(ninety_nine, sqrt).ok()?;
 	let ratio = I95F33::from_num(1_u8) - pow;
-	let inflation = I95F33::from_num(x) * ratio;
+	let inflation = I95F33::checked_from_num(x)? * ratio;
```
hackfisher, line 69: What will happen to the mark `?` here? return `None`?

What will happen if it returns `None` here?

hackfisher: What's the max value of  I95F33, it there possibility of overflow, just in case it was designed to utilise overflow

AurevoirXavier: Downstream usages:
- https://github.com/darwinia-network/darwinia/blob/9f3266b6af020edbce8c16b9ece688c7b79bad75/pallet/staking/src/lib.rs#L818-L826

hackfisher: > What's the max value of I95F33, it there possibility of overflow, just in case it was designed to utilise overflow

Max value is 19807040628566084398385987584, 1.98 * 10 ^28, max value of ring balance could be 100 * 10^8 * 10^18, for the case of ring, it is not likely to cause a result None here.

AurevoirXavier: See the comment here: https://github.com/darwinia-network/darwinia/blob/9f3266b6af020edbce8c16b9ece688c7b79bad75/core/inflation/src/lib.rs#L62

boundless-forest (approved):

PR #1292, opened 2023-10-09
OpenGov experiment

AurevoirXavier:


runtime/pangolin/src/pallets/governance/track.rs
```diff
@@ -0,0 +1,236 @@
...
+						Ok(*track_id)
+					} else {
+						Err(())
+					}
+				},
+				_ => Err(()),
+			}
+		} else if let Ok(custom_origin) = custom_origins::Origin::try_from(id.clone()) {
+			if let Some((track_id, _)) = Self::tracks().iter().find(|(_, track)| {
+				if let Ok(track_custom_origin) = custom_origins::Origin::from_str(track.name) {
```
boundless-forest, line 196: If the input is `GeneralAdmin` custom origin, no track_id can be matched following the logic here.

runtime/pangolin/src/pallets/governance/track.rs
```diff
@@ -0,0 +1,236 @@
...
+			name: "root",
+			// A limit for the number of referenda on this track that can be being decided at once.
+			// For Root origin this should generally be just one.
+			max_deciding: 5,
+			// Amount that must be placed on deposit before a decision can be made.
+			decision_deposit: DARWINIA_PROPOSAL_REQUIREMENT,
+			// Amount of time this must be submitted for before a decision can be made.
+			prepare_period: 2 * MINUTES,
+			// Amount of time that a decision may take to be approved prior to cancellation.
+			decision_period: 2 * MINUTES,
```
boundless-forest, line 46: I see that you have shorten lots of time config from 10 mins to 2 mins. After I did some local tests, the feedback is that 2 mins is too short, you can not even understand what happened, the time is out and proposal has been rejected.  So I suggest making the time value larger.

PR #1294, opened 2023-10-11
Redesign commission rule

AurevoirXavier:
Resolve part of https://github.com/darwinia-network/darwinia/discussions/1238.


AurevoirXavier:
cc @hackfisher @xiaoch05

pallet/staking/src/lib.rs
```diff
@@ -464,10 +458,6 @@ pub mod pallet {
 		pub fn collect(origin: OriginFor<T>, commission: Perbill) -> DispatchResult {
 			let who = ensure_signed(origin)?;
 
-			if commission > T::MaxCommission::get() {
-				Err(<Error<T>>::CommissionTooHigh)?;
-			}
-
 			<Collators<T>>::mutate(&who, |c| *c = Some(commission));
```
xiaoch05, line 461: Should we check the commission not bigger than 100% and less than 0%?

boundless-forest: No need to do that, the type `Perbill` type adds such restrictions to the input value. 

```sh
2023-10-12 13:45:49 panicked at 'Bad input data provided to query_info: Codec error', /home/bear/coding/rust-space/darwinia/runtime/pangolin/src/lib.rs:234:1    
2023-10-12 13:45:50 panicked at 'Bad input data provided to validate_transaction: Codec error', /home/bear/coding/rust-space/darwinia/runtime/pangolin/src/lib.rs:234:1 
```
You transaction with invalid commissions as you said will be rejected by the runtime.

pallet/staking/src/lib.rs
```diff
@@ -901,24 +891,24 @@ pub mod pallet {
...
-								Some(IndividualExposure { who: n, value: n_power })
+								Some(IndividualExposure { who: n, vote: nominator_v })
 							} else {
 								None
 							}
 						})
 						.collect();
 
-					((c, Exposure { total: t_power, nominators: i_exposures }), t_power)
+					((c, Exposure { vote: collator_v, nominators }), collator_v)
```
boundless-forest, line 911: One small suggestion for improvement: 
You can calculate the total_power first and then mul the `(Perbill::one() - cm)` later, in this way reducing the mul times.

pallet/staking/src/lib.rs
```diff
@@ -940,6 +930,7 @@ pub use pallet::*;
 
 type RewardPoint = u32;
 type Power = u32;
+type Vote = u32;
```
boundless-forest, line 933: Can you add some comment on the differences between the vote and power here? I am not sure if the vote equals to the  power * adjustment.

pallet/staking/src/lib.rs
AurevoirXavier: Nope, `nomonator_v` also need to be scaled. But I can take the `Perbill::one - cm` out of the loop.

PR #1310, opened 2023-11-01
Simplify Crab token system

AurevoirXavier:
Refer to #1273.

- Stop inflation on Crab and Pangolin
- Add `OnSessionEnd` to `Config`
- Remove `RingCurrency`, `UnixTime`, `RewardReminder` and `PayoutFraction` from `Config`
- Infation parameters need to be configured in runtime now


pallet/staking/src/lib.rs
```diff
@@ -867,14 +819,14 @@ pub mod pallet {
 					}
 				}
 
-				if T::RingCurrency::deposit_into_existing(&c, c_payout).is_ok() {
+				if T::OnSessionEnd::reward(&c, c_payout).is_ok() {
```
boundless-forest, line 822: Suggest adding another event to track the cases of reward failures. One of the reasons after this PR is that the treasury fund has run out. Catch the error in the pallet instead of the `reward` function to avoid duplicate error handling.

boundless-forest (commented):
I have some concerns about the payout  of the Treasury reward. I haven't seen any test cases to ensure that there is no calculation overflow or potential panic if the Treasury is unable to pay for the 20_000 * UNIT.  Others LGTM

runtime/crab/src/pallets/staking.rs
```diff
@@ -70,8 +72,25 @@ impl darwinia_staking::Stake for KtonStaking {
...
+			&Treasury::account_id(),
+			who,
+			amount,
+			frame_support::traits::ExistenceRequirement::KeepAlive,
+		) {
+			// TODO: log
+			// log::error!("[runtime::staking] reward error: {:?}", e);
+		}
+
+		Ok(())
```
boundless-forest, line 92: Currently, the reward here always returns `Ok()`, it seems problematic. If the transfer fails, the `n_payout` should not included in the `actual_payout`.

boundless-forest (dismissed):

boundless-forest (approved):

runtime/crab/src/pallets/staking.rs
```diff
@@ -70,8 +72,20 @@ impl darwinia_staking::Stake for KtonStaking {
 	}
 }
 
-frame_support::parameter_types! {
-	pub const PayoutFraction: sp_runtime::Perbill = sp_runtime::Perbill::from_percent(40);
+pub enum OnCrabSessionEnd {}
+impl darwinia_staking::OnSessionEnd<Runtime> for OnCrabSessionEnd {
+	fn calculate_reward(_maybe_inflation: Option<Balance>) -> Balance {
+		20_000 * UNIT
+	}
```
hackfisher, line 79: 20000 for 1 session( 6 hours) for Crab, so we can calculate the cost for collators staking:

2,920,000 CRAB block rewards for 1 year, much less compared to currently around 32,000,000 CRAB block rewards for 1 year?

runtime/crab/src/pallets/staking.rs
```diff
@@ -70,8 +72,20 @@ impl darwinia_staking::Stake for KtonStaking {
...
+
+	fn reward(who: &AccountId, amount: Balance) -> sp_runtime::DispatchResult {
+		<Balances as Currency<AccountId>>::transfer(
+			&Treasury::account_id(),
+			who,
+			amount,
+			frame_support::traits::ExistenceRequirement::KeepAlive,
+		)
```
hackfisher, line 87: What will happen if the treasury does not have enough balance for paying, will it call the session end fail without making any changes or stall block?

runtime/crab/src/pallets/staking.rs
```diff
@@ -70,8 +72,20 @@ impl darwinia_staking::Stake for KtonStaking {
 	}
 }
 
-frame_support::parameter_types! {
-	pub const PayoutFraction: sp_runtime::Perbill = sp_runtime::Perbill::from_percent(40);
+pub enum OnCrabSessionEnd {}
+impl darwinia_staking::OnSessionEnd<Runtime> for OnCrabSessionEnd {
```
hackfisher, line 76: `fn inflate()` is not implemented for OnCrabSessionEnd, so it will use default implementation?

hackfisher (commented):
LGTM

hackfisher (approved):

PR #1313, opened 2023-11-09
Improve staking payout performance

AurevoirXavier:


pallet/staking/src/lib.rs
```diff
@@ -181,18 +192,18 @@ pub mod pallet {
...
+	#[pallet::getter(fn previous_exposure_of)]
+	pub type PreviousExposures<T: Config> =
 		StorageMap<_, Twox64Concat, T::AccountId, Exposure<T::AccountId>>;
 
-	/// Next stakers' exposure.
+	/// Active stakers' exposure.
 	#[pallet::storage]
 	#[pallet::unbounded]
-	#[pallet::getter(fn next_exposure_of)]
-	pub type NextExposures<T: Config> =
```
hackfisher, line 195: NextExposures is for elect and store info related to next session, should be be deleted IMO

hackfisher (changes requested):
Migration is required in the runtime.

pallet/staking/src/lib.rs
```diff
@@ -735,215 +827,276 @@ pub mod pallet {
...
 
-				for n_exposure in c_exposure.nominators {
-					let n_payout =
-						Perbill::from_rational(n_exposure.vote, c_exposure.vote) * n_payout;
+			map.into_iter().for_each(|(c, p)| {
+				let r = Perbill::from_rational(p, sum) * amount;
 
-					if c == n_exposure.who {
-						// If the collator nominated themselves.
+				<Unpaid<T>>::mutate(&c, |u| *u = u.map(|u| u + r).or(Some(r)));
```
boundless-forest, line 904: Following the latest design, I think this line should be moved down to the `T::InflationManager::reward(&staking_pot, r).is_ok()` branch. If the inflation manager reward failture fails, the collator does not get any token reward. No unpaid.

pallet/staking/src/lib.rs
```diff
@@ -735,215 +827,276 @@ pub mod pallet {
...
 {
 	/// Generic session termination procedures.
 	fn on_session_end() {
 		let inflation = Self::inflate();
 		let reward = Self::calculate_reward(inflation);
-		let actual_payout = <Pallet<T>>::payout(reward);
+		let actual_reward = <Pallet<T>>::dispatch_session_reward(reward);
 
-		Self::clean(inflation.unwrap_or(reward).saturating_sub(actual_payout));
+		if inflation != 0 {
```
boundless-forest, line 1060: Suggest changing the condition to `if (inflation - actual_reward) > 0`. Much easier to read and if these two values are equal, no clear happens.

pallet/staking/src/lib.rs
```diff
@@ -735,215 +827,276 @@ pub mod pallet {
...
-		// - `total_ring_power = (amount / total_staked_ring) * HALF_POWER`
-		// - `total_kton_power = (amount / total_staked_kton) * HALF_POWER`
-		fn balance2power<P>(amount: Balance) -> Power
-		where
-			P: frame_support::StorageValue<Balance, Query = Balance>,
-		{
-			(Perquintill::from_rational(amount, P::get().max(1)) * 500_000_000_u128) as _
+		/// Calculate the power of the given account.
+		#[cfg(any(feature = "runtime-benchmarks", test))]
+		pub fn quick_power_of(who: &T::AccountId) -> Power {
```
boundless-forest, line 843: Why not:
```rs
#[cfg(any(feature = "runtime-benchmarks", test))]
pub fn quick_power_of(who: &T::AccountId) -> Power {
	let ring_pool = <RingPool<T>>::get();
	let kton_pool = <KtonPool<T>>::get();
	Self::power_of(who, ring_pool, kton_pool)
}
```

pallet/staking/src/lib.rs
```diff
@@ -735,215 +827,276 @@ pub mod pallet {
...
+}
 
-	fn clean(_unissued: Balance) {}
+/// Exposure cache's state.
+#[allow(missing_docs)]
+#[cfg_attr(any(feature = "runtime-benchmarks", feature = "try-runtime"), derive(PartialEq))]
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, RuntimeDebug)]
+pub enum ExposureCacheState {
+	Previous,
+	Current,
```
boundless-forest, line 1098: It seems that `Current` is not used?

pallet/staking/src/lib.rs
```diff
@@ -260,6 +340,17 @@ pub mod pallet {
...
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_: BlockNumberFor<T>) -> Weight {
+			call_on_exposure!(Previous::iter_keys()
+				// TODO?: make this value adjustable
+				.drain()
+				.take(1)
+				.fold(Zero::zero(), |acc, e| acc + Self::payout_inner(e).unwrap_or(Zero::zero())))
+			.unwrap_or_default()
+		}
```
boundless-forest, line 352:
```rs
#![allow(unused)]
fn main() {
    let mut v = vec![1, 2, 3, 4];
    let u: Vec<_> = v.drain(..).take(1).collect();
    assert_eq!(v, &[]);
    assert_eq!(u, &[1]);
}
```
The `drain` here will remove all items in the cache container. it's problematic.

pallet/staking/src/lib.rs
```diff
@@ -181,18 +211,62 @@ pub mod pallet {
 	#[pallet::getter(fn collator_of)]
 	pub type Collators<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, Perbill>;
 
-	/// Current stakers' exposure.
+	/// Exposure cache states.
+	///
+	/// To avoid extra DB RWs during new session, such as:
```
boundless-forest, line 216: Honestly, this is indeed a great solution to avoid copying the current session's callator information into another container. However, according to my understanding of this problem, ExposureCacheState is likely a pointer to either the current session storage in the elect() or the previous session storage in the hook, it switches as the session switches. So, I think two pointers are enough, now there are three.

pallet/staking/src/lib.rs
```diff
@@ -735,215 +827,276 @@ pub mod pallet {
...
+					((c, Exposure { commission: cm, vote: collator_v, nominators }), collator_v)
 				})
 				.collect::<Vec<_>>();
 
 			collators.sort_by(|(_, a), (_, b)| b.cmp(a));
 
+			let cache_states = <ExposureCacheStates<T>>::get();
+
+			#[allow(deprecated)]
+			call_on_exposure!(cache_states, Next::remove_all(None))?;
```
boundless-forest, line 1028: Suggest moving this out to the `prepare_new_session` after `shift_exposure_cache_states`

pallet/staking/src/lib.rs
AurevoirXavier: Store current exposures to reward the current exposures in new session

pallet/staking/src/lib.rs
AurevoirXavier: Rust `drain` is different from Substrate `drain`.

But all iterators are lazy.

You can try these:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6b088f2f4942e6982803d607d07f453b

```rs
#[test]
fn playground() {
	ExtBuilder::default().collator_count(2).genesis_collator().build().execute_with(|| {
		dbg!(call_on_exposure_test!(Next::iter_keys().collect::<Vec<_>>()));
		call_on_exposure_test!(Next::iter_keys().drain().take(1).collect::<Vec<_>>());
		dbg!(call_on_exposure_test!(Next::iter_keys().collect::<Vec<_>>()));
	});
}
```

pallet/staking/src/lib.rs
AurevoirXavier: Definitely. I didn't notice this before.

pallet/staking/src/lib.rs
AurevoirXavier: How to implement this with two pointers?

pallet/staking/src/lib.rs
AurevoirXavier: Renaming is acceptable, but it should still be counted even if the reward fails.

boundless-forest: >  it should still be counted even if the reward fails.

In the case of reward fails, the actual reward that the collator receives is less than the PendingReward value, in the next session, the collator may lose their original fund to pay out the nominator's reward. Is this expected behaviour?

runtime/crab/src/weights/darwinia_staking.rs
```diff
@@ -199,4 +199,7 @@ impl<T: frame_system::Config> darwinia_staking::WeightInfo for WeightInfo<T> {
 			.saturating_add(Weight::from_parts(0, 0))
 			.saturating_add(T::DbWeight::get().writes(1))
 	}
+	fn payout() -> Weight {
+		Default::default()
+	}
```
boundless-forest, line 204: Update benchmark result later?

pallet/staking/src/lib.rs
AurevoirXavier: Yep

hackfisher (approved):

runtime/crab/src/weights/darwinia_staking.rs
AurevoirXavier: Will do this in the release PR

boundless-forest (approved):

PR #1336, opened 2023-12-04
Enable maintenance pallets for the testnet

boundless-forest:
This PR introduces `pallet-tx-pause` to the Pangolin network,  one of the two maintenance pallets in the polkadot-sdk upstream repo. It allows to pause some dispatch calls in the case of emergency.  It's worth noting the reviewers is the two origins and make sure they are secure enough before going into the production. As for why don't add `pallet-safe-mode`? Two reasons:
1. In `safe-mod`, the `enter` call is permissionless, that means anyone who has enough fund over the Limitation can put the system into safe-mode state. I don't think this is a secure enough solution for our system now. If the max deposit value is not set to an appropriate value, your system is at risk.
2. Based on the maintenance experience we've had so far, `tx-pause` is enough IMO. The most common senerios is that some dispatch calls are vulnerable, pausing these  calls is enough.


runtime/common/src/test.rs
```diff
@@ -848,3 +848,98 @@ macro_rules! impl_messages_bridge_tests {
...
+						RuntimeOrigin::root(),
+						full_name(b"Balances", b"transfer")
+					));
+
+					assert_err!(
+						TxPause::pause(
+							RuntimeOrigin::signed(H160::default().into()),
+							full_name(b"Balances", b"transfer")
+						),
+						DispatchError::BadOrigin
```
AurevoirXavier, line 884: Add transfer testing after the pause.

runtime/pangolin/src/pallets/tx_pause.rs
```diff
@@ -0,0 +1,44 @@
...
+use crate::*;
+
+/// Calls that cannot be paused by the tx-pause pallet.
+pub struct TxPauseWhitelistedCalls;
+impl frame_support::traits::Contains<pallet_tx_pause::RuntimeCallNameOf<Runtime>>
+	for TxPauseWhitelistedCalls
+{
+	fn contains(full_name: &pallet_tx_pause::RuntimeCallNameOf<Runtime>) -> bool {
+		match (full_name.0.as_slice(), full_name.1.as_slice()) {
+			(b"System", b"remark_with_event") => true,
```
AurevoirXavier, line 29: Any specific reason?

AurevoirXavier: Yep, I know those tests. They are written in common runtime, which means that all runtimes need to be configured accordingly.

AurevoirXavier: That could potentially become a problem in the future. Honestly, I'm not sure if we should whitelist that call for the mainnet. Can we assume some calls are "unconcern" about security.

PR #1381, opened 2024-01-12
Add `tx-pause` to all runtimes

AurevoirXavier:


boundless-forest (commented):
Please add `darwinia_common_runtime::impl_maintenance_tests! {}` to the test file under each runtime.

Issue #1391, opened 2024-01-19
Clear council, democracy locks for users after removing governance v1


AurevoirXavier:
We can call this before the upgrade.

https://github.com/paritytech/polkadot-sdk/blob/9db92115d023689581e8cf49edb9500b0dcced53/substrate/frame/democracy/src/lib.rs#L959

AurevoirXavier:
### Remaining locks
```json
[
  {
    "publicKey": "0xd891ce6a97b4f01a8b9b36d0298aa3631fe2eef5",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0xabcf7060a68f62624f7569ada9d78b5a5db0782a",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0x88a39b052d477cfde47600a7c9950a441ce61cb4",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0x9f33a4809aa708d7a399fedba514e0a0d15efa85",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0x0a1287977578f888bdc1c7627781af1cc000e6ab",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0xe59261f6d4088bcd69985a3d369ff14cc54ef1e5",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0x7ae2a0914db8bfbdad538b0eac3fa473a0e07843",
    "lockId": "democrac"
  },
  {
    "publicKey": "0xacfa39b864e42d1bd3792783a571d2958af0bf1f",
    "lockId": "democrac"
  },
  {
    "publicKey": "0x3e25247cff03f99a7d83b28f207112234fee73a6",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0xb2960e11b253c107f973cd778bbe1520e35e8602",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0x4ed7ae57608cf4f60753cde4f49cf821c293ed2a",
    "lockId": "democrac"
  },
  {
    "publicKey": "0x5b7544b3f6abd9e03fba494796b1ee6f9543e2e4",
    "lockId": "phrelect"
  },
  {
    "publicKey": "0x44cda595218ddb3810fb66c2e982f50ea00255ee",
    "lockId": "phrelect"
  }
]
```

AurevoirXavier:
### Darwinia Remaining Locks (Ignore Democracy for Current Version)
```json
[
  {
[...]

PR #1398, opened 2024-02-02
Remove deprecated locks

AurevoirXavier:
Part of #1391.

### DIFF
--- a/runtime/crab/src/migration.rs
+++ b/runtime/crab/src/migration.rs
@@ -41,47 +41,33 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 }
 
 fn migrate() -> frame_support::weights::Weight {
-	let items: &[(&[u8], &[&[u8]])] = &[
-		(b"MessageGadget", &[b"CommitmentContract"]),
-		(
-			b"EcdsaAuthority",
-			&[
-				b"Authorities",
-				b"NextAuthorities",
-				b"Nonce",
-				b"AuthoritiesChangeToSign",
-				b"MessageRootToSign",
-			],
-		),
-		(
-			b"Council",
-			&[b"Proposals", b"ProposalOf", b"Voting", b"ProposalCount", b"Members", b"Prime"],
-		),
-		(
-			b"Democracy",
-			&[
-				b"PublicPropCount",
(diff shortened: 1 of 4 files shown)


runtime/crab/src/migration.rs
```diff
@@ -41,47 +41,33 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
...
+		if let Ok(acct) = array_bytes::hex_n_into::<_, AccountId, 20>(acct) {
+			<Locks<Runtime>>::mutate(acct, |ls| {
+				ls.retain(|l| &l.id != *lid);
+			});
+		}
 	});
 
 	// frame_support::weights::Weight::zero()
-	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(0, w)
+	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(0, 26)
```
boundless-forest, line 72: Can you add some note/comments about the concrete read or write operation here? Seems two writes for single item.

runtime/darwinia/src/migration.rs
```diff
@@ -42,20 +42,27 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
...
+		if let Ok(acct) = array_bytes::hex_n_into::<_, AccountId, 20>(acct) {
+			<Locks<Runtime>>::mutate(acct, |ls| {
+				ls.retain(|l| &l.id != *lid);
+			});
+		}
+	});
 
 	// frame_support::weights::Weight::zero()
-	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(0, 5)
+	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(0, 16)
```
boundless-forest, line 67: The same as the above comment.

runtime/crab/src/migration.rs
AurevoirXavier: `Balance::Locks + System::Account`

AurevoirXavier: Simply `Balances::Lock::writes() * 2`. 

It's kind of arbitrated choice.
Leave some redundant space.

boundless-forest (approved):

Issue #1405, opened 2024-02-11
Generalize Interoperate between Substrate and EVM smart contracts


hackfisher:
1. After #1404 and #1382 been processed, the pallet `message_transact`  can be renamed with an generalized pallet for handling interoperate from substrate root and custom origins to EVM:

https://github.com/darwinia-network/darwinia/blob/502f39d4dd04bbaa785c4cd3b05a57792f2f964d/pallet/message-transact/src/lib.rs#L124-L127

Root and custom origins will transact as configured special H160 Ethereum addresses, such as b"root", require no duplicate with any smartcontract and EOA, this will work for most cases, if there are extra data in the origin, they will be passed to the Ethereum transaction in calldata.

The traits of the NonEthereumOriginTransact Pallet could be similar to something like:

```
// the origin must match with root origin or custom origins
// can be called by XCMP or OpenGov proposals, uses cases includes managing ERC20 tokens etc in Treasury or do xToken related txs.
pub fn transact_in_evm(
    origin: OriginFor<T>,
    mut transaction: Box<Transaction>,
) -> DispatchResultWithPostInfo
```

2. Root and custom origins can be configured to be dispatched by runtime-configured specific smart contract. By replacing `type AddressMapping = pallet_evm::IdentityAddressMapping;`

https://github.com/darwinia-network/darwinia/blob/502f39d4dd04bbaa785c4cd3b05a57792f2f964d/runtime/darwinia/src/pallets/evm.rs#L162

with customized `AddressMapping` considering the root and customs origins for specific smart contract.

hackfisher:
Refer: https://github.com/darwinia-network/darwinia/issues/1282

hackfisher:
Review the usage to setting msg.value here, any risks?

Will it still work if the msg.sender do not have balance > msg.value? Who will pay the gas? (runtime could be in init()), in this way no one will pay gas, its free?

PR #1421, opened 2024-03-01
Fix staking reward distribute issue

boundless-forest:
This PR fixes the staking distribution errors by removing the chain id validation, it's okay for me because of the runtime trasact can only be used at runtime. It also involves lots of renaming staff, which is discussed in the PR discussion. No pallet storage in this pallet, no migration.

The `test_legacy_transaction_with_sufficient_balance` test case illustrates how to calculate the amount of fee that goes to the sender. @AurevoirXavier


hackfisher (commented):
Since we've removed S2S, can we rename `pub fn message_transact(` and define a new Origin for dispatch?

https://github.com/darwinia-network/darwinia/issues/1405

pallet/staking/src/lib.rs
```diff
@@ -1315,44 +1313,42 @@ where
...
+				54, 178, 65, 176, 97, 163, 106, 50, 171, 127, 232, 108, 122, 169, 235, 89, 45, 213,
+				144, 24, 205, 4, 67, 173, 192, 144, 53, 144, 193, 107, 2, 176,
+			]),
 		};
 		// b"sc/ktstk"
 		let sender =
 			H160([115, 99, 47, 107, 116, 115, 116, 107, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]);
 
 		if let Err(e) = <darwinia_message_transact::Pallet<T>>::message_transact(
 			LcmpEthOrigin::MessageTransact(sender).into(),
```
hackfisher, line 1348:
```suggestion
			Sub2EthOrigin::MessageTransact(sender).into(),
```

boundless-forest:
> Since we've removed S2S, can we rename `pub fn message_transact(` and define a new Origin for dispatch?
> 
> #1405

It make a lot of sense. I didn't come up with a better rename yet~ any suggestions?

pallet/staking/src/lib.rs
boundless-forest: - pallet name: `RuntimeTransact`
- origin name: `RuntimeEthOrigin`
- function name: `runtime_transact(xxx)`

What do you think of the renaming above? 

`sub2`, the substrate is renamed to `polkadot-vm`, so I think we should avoid introducing something like `subxxx` anymore ~ `runtime` is a better word for me. The meaning is also clear enough.

hackfisher: LGTM

AurevoirXavier:
I feel the word 'runtime' in the name is redundant.

boundless-forest:
> I feel the word 'runtime' in the name is redundant.

The `runtime` here mainly want to differ the evm transaction from the RPC transactions.

Hmm, what do you suggest? If you have, comment the renaming list like this https://github.com/darwinia-network/darwinia/pull/1421#discussion_r1508816676

pallet/runtime-transact/src/lib.rs
```diff
@@ -44,23 +44,23 @@ use sp_std::boxed::Box;
 pub use pallet::*;
 
 #[derive(Clone, Eq, PartialEq, RuntimeDebug, Encode, Decode, MaxEncodedLen, TypeInfo)]
-pub enum LcmpEthOrigin {
-	MessageTransact(H160),
+pub enum RuntimeEthOrigin {
```
AurevoirXavier, line 47: Why use enum here

AurevoirXavier:
> Hmm, what do you suggest? If you have, comment the renaming list like this https://github.com/darwinia-network/darwinia/pull/1421#discussion_r1508816676

EvmTxWrapper

EvmTxOrigin

fn send/transact

pallet/runtime-transact/src/lib.rs
boundless-forest: Currently only stake module uses it, it may later contain several variables, such as `StakingTransact(H160)`, `PalletBTransact(H160)`,xxx.

boundless-forest:
> EvmTxWrapper & EvmTxOrigin

I don't think this is a good name in this scenario, it's much more about the ethereum transaction instead of evm~.

> fn send/transact

This makes sense to me.

AurevoirXavier:
> I don't think this is a good name in this scenario, it's much more about the ethereum transaction instead of evm~.

Then RuntimeTransact looks even more weird. 

I don't think the key point is ethereum. It's more about how do we adapt evm transaction in substrate runtime.

boundless-forest:
> It's more about how do we adapt/construct evm transaction within substrate runtime context.

I think this sentence is partly true. "evm transaction" is a bit of a misleading term, there is no transaction concept for the evm, it's just an executor or runner, used to execute the `ethereum` transaction. Based on the discusstion, I think maybe `RuntimeEthereum` would be a better name.

hackfisher (commented):
For the naming, I think runtime transact looks good for me because this indicates the source of this tx is from called from runtime to EVM executor, distinguishing it from ordinary evm transact using ecdsa unsigned extrinsic. Alternative naming would better to find a replacement for the runtime, `internal`, `system`, `private`?

Did we make test pass on the Pangolin yet?

hackfisher (approved):

PR #1431, opened 2024-03-05
Add migration

AurevoirXavier:
The main task is to remove the S2S storage. Additionally, this resolves issue #1391.

### DIFF
--- a/runtime/common/src/migration_helper.rs
+++ b/runtime/common/src/migration_helper.rs
@@ -1,0 +1,31 @@
+// substrate
+pub use frame_support::migration;
+
+/// Pallet migration helper.
+pub struct PalletCleaner {
+	pub name: &'static [u8],
+	pub values: &'static [&'static [u8]],
+	pub maps: &'static [&'static [u8]],
+}
+impl PalletCleaner {
+	/// Remove all storage from a pallet.
+	pub fn remove_all(&self) -> u32 {
+		self.remove_storage_values() + self.remove_storage_maps()
+	}
+
+	/// Remove multiple storage value from a pallet at once.
+	pub fn remove_storage_values(&self) -> u32 {
+		self.values.iter().for_each(|i| {
+			let _ = migration::clear_storage_prefix(self.name, i, &[], None, None);
+		});
+
+		self.values.len() as u32
+	}
+
+	/// Remove multiple storage map from a pallet at once.
+	pub fn remove_storage_maps(&self) -> u32 {
+		self.maps.iter().fold(0, |a, i| {
+			a + migration::clear_storage_prefix(self.name, i, &[], None, None).backend
+		})
+	}
+}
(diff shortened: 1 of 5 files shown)


runtime/crab/src/migration.rs
```diff
@@ -68,6 +68,52 @@ fn migrate() -> frame_support::weights::Weight {
...
+			b"PalletOwner",
+			b"PalletOperatingMode",
+		],
+		maps: &[b"ImportedHashes", b"ImportedHeaders"],
+	}
+	.remove_all();
+	w += migration_helper::PalletCleaner {
+		name: b"BridgePolkadotGrandpa",
+		values: &[b"PalletOwner", b"PalletOperatingMode"],
+		maps: &[b"ParasInfo", b"ImportedParaHeads", b"ImportedParaHashes"],
```
boundless-forest, line 90: The migration value and maps seems differ from the grandpa pallet storage layout.

runtime/common/src/migration_helper.rs
```diff
@@ -0,0 +1,31 @@
...
+		self.remove_storage_values() + self.remove_storage_maps()
+	}
+
+	/// Remove multiple storage value from a pallet at once.
+	pub fn remove_storage_values(&self) -> u32 {
+		self.values.iter().for_each(|i| {
+			let _ = migration::clear_storage_prefix(self.name, i, &[], None, None);
+		});
+
+		self.values.len() as u32
```
boundless-forest, line 22: Are you sure that the write times for the clearing are the same as the `self.values.len()`? For example, If there is a pallet storage map with 100 elements, the write times is only 1 instead of 100?

runtime/darwinia/src/migration.rs
```diff
@@ -72,6 +79,52 @@ fn migrate() -> frame_support::weights::Weight {
...
+			b"CollateralSlashProtect",
+			b"AssignedRelayersNumber",
+		],
+		maps: &[b"Orders", b"RelayersMap"],
+	}
+	.remove_all();
+
 	// frame_support::weights::Weight::zero()
-	RuntimeBlockWeights::get().max_block
+	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(0, w as _)
```
boundless-forest, line 129: I suggest changes this to the `max_block`, the currently mannual `w` missing the write operation for the `create_account` above. Also apply to the other runtimes.

runtime/crab/src/migration.rs
```diff
@@ -68,6 +68,52 @@ fn migrate() -> frame_support::weights::Weight {
 		}
 	});
 
+	let mut w = 26;
+
+	w += migration_helper::PalletCleaner {
+		name: b"BridgeKusamaGrandpa",
```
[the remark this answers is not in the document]
boundless-forest: Solved in another PR.

runtime/common/src/migration_helper.rs
boundless-forest: Hmm, this is `remove_storage_values`, it makes sense that way.

PR #1434, opened 2024-03-05
Fix migration

hackfisher:


runtime/darwinia/src/migration.rs
```diff
@@ -107,12 +107,6 @@ fn migrate() -> frame_support::weights::Weight {
 		maps: &[b"InboundLanes", b"OutboundLanes", b"OutboundMessages"],
 	}
 	.remove_all();
-	w += migration_helper::PalletCleaner {
```
AurevoirXavier, line 110: It looks like there are some storage items under the message pallet.

hackfisher: Yes, but it seems that this is duplicate, it has already been done in previous lines.

PR #1449, opened 2024-03-12
Fix release notifier

AurevoirXavier:


.github/workflows/notify-release.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           curl -XPOST \
             -H "Authorization:Bearer ${{ secrets.GH_TKN_DARWINIA }}" \
             -H "Accept:application/vnd.github+json" \
-            "https://api.github.com/repos/darwinia-network/wasm-runtime-overrides/dispatches" \
+            "https://api.github.com/repos/darwinia-network/darwinia-releases/dispatches" \
```
fewensa, line 21: It's looks wrong, we changed use workflow_dispatchs, the documents here https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event

May we should change to 

```
"https://api.github.com/repos/darwinia-network/darwinia-releases/actions/workflows/wasm.yaml/dispatches" \
-d '{"ref":"main","inputs":{"network":"mainnet","tag":"'${TAG}'"}}'
```

.github/workflows/notify-release.yml
```diff
@@ -32,5 +32,5 @@ jobs:
...
```
fewensa, line 35: Same here

fewensa (changes requested):

.github/workflows/notify-release.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           curl -XPOST \
```
fewensa, line 18:
```suggestion
          curl -X POST \
```

fewensa (approved):

PR #1463, opened 2024-04-01
Half the gas price

### LINKED ISSUE #1461: Support for widely used create2 factory on Darwinia and Pangolin
## Repo
https://github.com/Arachnid/deterministic-deployment-proxy

## Issue
Due to gas prices exceeding 100 gwei on the Darwinia and Pangolin networks for an extended period, the widely used create2 factory that does not require a key (keyless)  cannot be deployed.

## Reference
- [foundry](https://github.com/foundry-rs/foundry/blob/617dfc28cb8206a0003edcf73a6f1058adaef740/crates/evm/core/src/backend/error.rs#L44)
- [hardhat](https://github.com/wighawag/hardhat-deploy/blob/e3727896a812fe1a17e04f52a4aecc42c7b13ed4/src/DeploymentsManager.ts#L594)
- [Sphinx](https://github.com/sphinx-labs/sphinx/blob/85f13b17ac4699cee128b26e771d33775f60557a/packages/contracts/src/constants.ts#L39)

boundless-forest:
Fix #1461 

Half the gas price after some internal discussion, The expected gas price should be 75GWEI after this PR.

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -84,9 +84,8 @@ macro_rules! impl_self_contained_call {
 				len: usize,
 			) -> Option<sp_runtime::transaction_validity::TransactionValidity> {
 				match self {
-					RuntimeCall::Ethereum(call) => {
-						call.validate_self_contained(info, dispatch_info, len)
-					},
+					RuntimeCall::Ethereum(call) =>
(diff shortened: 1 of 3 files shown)


AurevoirXavier:
Please don't modify the toolchain version.

PR #1475, opened 2024-04-15
Implement DIP-6

AurevoirXavier:


node/src/chain_spec/crab.rs
```diff
@@ -200,6 +200,7 @@ pub fn genesis_config() -> ChainSpec {
 				darwinia_staking: DarwiniaStakingConfig {
 					now: SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis(),
 					elapsed_time: 0,
+					max_unstake_ring: 10_000_000 * UNIT,
```
hackfisher, line 203: What's the max deposit currently?

```suggestion
					max_unstake_ring: 20_000_000 * UNIT,
```

node/src/chain_spec/crab.rs
```diff
@@ -303,6 +304,7 @@ fn testnet_genesis(
 		darwinia_staking: DarwiniaStakingConfig {
 			now: SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_millis(),
 			elapsed_time: 0,
+			max_unstake_ring: 10_000_000 * UNIT,
```
hackfisher, line 307:
```suggestion
			max_unstake_ring: 20_000_000 * UNIT,
```

pallet/staking/src/lib.rs
```diff
@@ -373,18 +382,13 @@ pub mod pallet {
 				} else {
 					<frame_system::Pallet<T>>::inc_consumers(&who)?;
 
-					*l = Some(Ledger {
-						staked_ring: Default::default(),
-						staked_deposits: Default::default(),
-						unstaking_ring: Default::default(),
-						unstaking_deposits: Default::default(),
-					});
+					*l = Some(Ledger { ring: Default::default(), deposits: Default::default() });
```
hackfisher, line 385: Also need rate limiting when stake

pallet/staking/src/lib.rs
```diff
@@ -413,64 +417,42 @@ pub mod pallet {
...
-
-		/// Claim the stakes from the pallet/contract account.
-		#[pallet::call_index(3)]
-		#[pallet::weight(<T as Config>::WeightInfo::claim())]
-		pub fn claim(origin: OriginFor<T>) -> DispatchResult {
-			let who = ensure_signed(origin)?;
+			if acc <= <MaxUnstakeRing<T>>::get() {
+				<AccumulateUnstake<T>>::put(acc);
+			} else {
+				Err(<Error<T>>::ExceedMaxUnstakeAmount)?;
```
hackfisher, line 453: Just want to make sure:

This is transactional right? The state changes will revert when meet `Err` here?

precompile/staking/src/lib.rs
```diff
@@ -93,38 +93,6 @@ where
...
-		let origin = handle.context().caller.into();
-
-		RuntimeHelper::<Runtime>::try_dispatch(
-			handle,
-			Some(origin).into(),
-			darwinia_staking::Call::<Runtime>::claim {},
-		)?;
-		Ok(true)
-	}
-
```
hackfisher, line 127: @JayJay1024 Precompile API break change Here

pallet/staking/src/lib.rs
AurevoirXavier: Yep, check the *assert no op* tests.

pallet/staking/src/lib.rs
AurevoirXavier: Ok, will set up a separate storage to track incoming stakes.

hackfisher: I think it is OK to use the same config

AurevoirXavier: Sure, I just came up with an idea.

```rs
enum Accumulator {
    Pos(u128),
    Neg(u128),
}
```

Stake will reduce the negative part, then increase the positive part.
The opposite is true for unstaking.

AurevoirXavier: Do not allow the positive or negative part to exceed 10,000,000,000, totaling 20,000,000,000 as per your request.

hackfisher: > Sure, I just came up with an idea.
> 
> 
> 
> ```rs
> 
> enum Accumulator {
> 
>     Pos(u128),
> 
>     Neg(u128),
> 
> }
> 
> ```
> 
> 
> 
> Stake will reduce the negative part, then increase the positive part.
> 
> The opposite is true for unstaking.

This make more outgoing unstake can be done by incoming stake.

Not quite sure about it, looks more complex for rate limiting.

AurevoirXavier: This seems more like a dynamic rate limiting strategy. It does introduce a bit more extra work, but that's acceptable to me.

Which one do you prefer?

hackfisher: It looks good to me after rethinking.

hackfisher: I still prefer simple one using same config, but dynamic one looks good to me, you decide it.

pallet/staking/src/lib.rs
hackfisher: OK, cool, I forgot transnational is default now.

runtime/pangolin/src/weights/darwinia_staking.rs
```diff
@@ -203,6 +168,9 @@ impl<T: frame_system::Config> darwinia_staking::WeightInfo for WeightInfo<T> {
 			.saturating_add(T::DbWeight::get().reads(4))
 			.saturating_add(T::DbWeight::get().writes(2))
 	}
+	fn set_max_unstake_ring() -> Weight {
+		Default::default()
+	}
```
boundless-forest, line 173: Remember to update the weight info here. I am fine with other changes.

boundless-forest (approved):

pallet/staking/src/lib.rs
```diff
@@ -946,6 +823,57 @@ where
...
+			Self::Neg(v) =>
+				if v >= amount {
+					Some(Self::Neg(v - amount))
+				} else {
+					let v = amount - v;
+
+					if v <= limit {
+						Some(Self::Pos(v))
+					} else {
+						None
```
hackfisher, line 847: Do we need to update self here, as there are side effects under other conditions?

hackfisher (approved):

PR #1485, opened 2024-04-25
Fix system account frozen field

### LINKED ISSUE #1483: Review and clear accounts that have non-zero`frozen` field
- Why those accounts have non-zero`frozen` field ? [Example](https://darwinia.subscan.io/account/0x7aE2a0914db8bFBdad538b0eAc3Fa473A0e07843)
- Clear if they are not needed anymore.

AurevoirXavier:
Closes #1483.

I'll test Darwinia again once the 6620 runtime upgrade is completed. Do not merge for now.

The algorithm is based on the code found at https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/balances/src/lib.rs#L1028-L1043.

### DIFF
--- a/runtime/darwinia/src/migration.rs
+++ b/runtime/darwinia/src/migration.rs
@@ -29,28 +29,12 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 	fn pre_upgrade() -> Result<Vec<u8>, sp_runtime::DispatchError> {
 		log::info!("pre");
 
-		assert!(migration::have_storage_value(b"EcdsaAuthority", b"Authorities", &[]));
-
-		<pallet_balances::Locks<Runtime>>::iter().for_each(|(k, v)| {
-			log::info!("{k:?}");
-			log::info!("{v:?}");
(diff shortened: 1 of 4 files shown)


hackfisher (approved):

runtime/darwinia/src/migration.rs
```diff
@@ -61,79 +45,42 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
...
-	.remove_all();
-
-	let _ = migration::clear_storage_prefix(
-		b"BridgeKusamaGrandpa",
-		b"ImportedHeaders",
-		&[],
-		Some(100),
-		None,
-	);
-
```
hackfisher, line 136: This may need to be kept

hackfisher (changes requested):

runtime/darwinia/src/migration.rs
AurevoirXavier: 300 keys remaining.

hackfisher (approved):

PR #1486, opened 2024-04-26
[Don't Merge] Correct `best_hash` for the frontier_backend

boundless-forest:


boundless-forest:
Close it and wait for the next Frontier upgrade.

PR #1511, opened 2024-06-05
Deprecate precompile

### LINKED ISSUE #1501: Deprecated BLS precompile
We don't need that anymore.

boundless-forest:
Fix #1501


runtime/koi/src/pallets/evm.rs
```diff
@@ -144,8 +144,7 @@ where
 			a if a == addr(0x602) =>
 				Some(<pallet_evm_precompile_conviction_voting::ConvictionVotingPrecompile<Runtime>>::execute(handle)),
 			// [0x800..) reserved for the experimental precompiles.
-			a if a == addr(0x800) =>
-				Some(<darwinia_precompile_bls12_381::BLS12381<Runtime>>::execute(handle)),
+			a if a == addr(0x800) => Some(Err(precompile_utils::prelude::revert("This precompile is no longer supported."))),
```
AurevoirXavier, line 147: What if return `None` here?

boundless-forest: Returning None means that the precompile does not exist. But in our case, we want to remind the users that this is a deprecated precompile.

AurevoirXavier: It indeed does not exist in Koi.

boundless-forest: Yes, it doesn't exist on the testnet(mainnet) anymore, but some applications or developers may still rely on this precompile. We need to avoid the situation where an application uses this precompile(address) now, but the precompile to this address changes to a another one later.  This will break the application's logic if it is not aware of the change.  

The Koi is the testnet for other networks, we should keep it the same behaviour across all the networks, otherwise, some application will work on the testnet but not on the mainnet.

PR #1516, opened 2024-06-21
To `polkadot-v1.7.2`

AurevoirXavier:


precompile/deposit/src/lib.rs
```diff
@@ -17,7 +17,6 @@
 // along with Darwinia. If not, see <https://www.gnu.org/licenses/>.
 
 #![cfg_attr(not(feature = "std"), no_std)]
-#![deny(unused_crate_dependencies)]
```
boundless-forest, line 20: Why remove this rule for these crates?

runtime/common/src/benchmark_helper.rs
```diff
@@ -0,0 +1,28 @@
+// crates.io
```
boundless-forest, line 1: Miss license header.

runtime/common/src/lib.rs
```diff
@@ -170,37 +169,40 @@ impl WeightToFeeT for WeightToFee {
...
-	fn polynomial() -> WeightToFeeCoefficients<Self::Balance> {
-		// Map base extrinsic weight to 1/200 UNIT.
+	fn polynomial() -> frame_support::weights::WeightToFeeCoefficients<Self::Balance> {
+		// Map base extrinsic weight to 1/800 UNIT.
 		let p = UNIT;
-		let q = 200 * Balance::from(ExtrinsicBaseWeight::get().ref_time());
+		let q = 800
+			* Balance::from(
+				frame_support::weights::constants::ExtrinsicBaseWeight::get().ref_time(),
+			);
```
boundless-forest, line 181: I remembered that when we update this value, the corresponding gas price will also change, but the gas_price tests don't seem to fire, weird.

runtime/common/src/pallet_config.rs
```diff
@@ -1,22 +1,83 @@
...
+use sp_runtime::traits::AccountIdConversion;
 use sp_std::prelude::*;
 
+/// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
+/// used to limit the maximal weight of a single extrinsic.
+const AVERAGE_ON_INITIALIZE_RATIO: sp_runtime::Perbill = sp_runtime::Perbill::from_percent(5);
+/// We allow `Normal` extrinsics to fill up the block up to 75%, the rest can be used by
+/// `Operational` extrinsics.
+pub const NORMAL_DISPATCH_RATIO: sp_runtime::Perbill = sp_runtime::Perbill::from_percent(75);
+/// We allow for 0.5 of a second of compute with a 12 second average block time.
```
boundless-forest, line 14: The current block time is 6s, please confirm this setting or update the comment.

precompile/deposit/src/lib.rs
AurevoirXavier: While compiling to WASM, it complains about `unused crate std, alloc`.

runtime/common/src/lib.rs
AurevoirXavier:
```diff
+ assert_eq!(pallet_config::WeightPerGas::get().ref_time(), 75000);
```

runtime/darwinia/src/lib.rs
```diff
@@ -151,32 +162,26 @@ frame_support::construct_runtime! {
 		Ethereum: pallet_ethereum = 36,
 		EVM: pallet_evm = 37,
 		EthTxForwarder: darwinia_ethtx_forwarder = 38,
-
-		// // Darwinia <> Crab
-		// BridgeKusamaGrandpa: pallet_bridge_grandpa::<Instance1> = 39,
-		// BridgeKusamaParachain: pallet_bridge_parachains::<Instance1> = 40,
-		// BridgeCrabMessages: pallet_bridge_messages::<Instance1> = 41,
-		// BridgeCrabDispatch: pallet_bridge_dispatch::<Instance1> = 42,
-		// CrabFeeMarket: pallet_fee_market::<Instance1> = 43
```
boundless-forest, line 160: Why not keep these comment and give `pallet_message_queue` a new index 44? 

Better safe than sorry, even if the storage of these modules hava been cleared.

runtime/crab/src/lib.rs
```diff
@@ -142,36 +152,31 @@ frame_support::construct_runtime! {
...
 		Ethereum: pallet_ethereum = 36,
 		EVM: pallet_evm = 37,
 		EthTxForwarder: darwinia_ethtx_forwarder = 38,
-
-		// // Crab <> Darwinia
-		// BridgePolkadotGrandpa: pallet_bridge_grandpa::<Instance1> = 39,
-		// BridgePolkadotParachain: pallet_bridge_parachains::<Instance1> = 40,
-		// BridgeDarwiniaMessages: pallet_bridge_messages::<Instance1> = 41,
-		// BridgeDarwiniaDispatch: pallet_bridge_dispatch::<Instance1> = 42,
-		// DarwiniaFeeMarket: pallet_fee_market::<Instance1> = 43
```
boundless-forest, line 156: The same as the darwinia runtime.

pallet/account-migration/src/lib.rs
```diff
@@ -435,34 +418,6 @@ pub mod pallet {
...
-				for r in rs.iter_mut().flatten() {
-					if r.account.0 == <AccountId32 as AsRef<[u8; 32]>>::as_ref(from)[..20] {
-						r.account = *to;
-
-						break;
-					}
-				}
-
-				migration::put_storage_value(b"Identity", b"Registrars", &[], rs);
-			}
```
boundless-forest, line 465: How do you deal with this identity in this pallet and the pallet-identity? Keep it for a while and then migrate to another solution?

Do you remember the way we deal with the lock balance about the identity in the account migration? We may need to take that into account.

runtime/common/src/lib.rs
boundless-forest: The increase in g here reduces the weight fee by 4 times. The increase in the block weight limit makes the WeightPerGas 4 times bigger. Ultimately, the gas price remains equal to the original one.

boundless-forest (commented):
It's a super huge pull request, it took me the whole afternoon to review the changes. I leave some feedback. Most of the changes look good to me. You removed two pallets, and there are also some tasks left to do or migrations needed, such as the asset description migration for the Koi testnet. I assume there will be another PR to handle the migration.

pallet/account-migration/src/lib.rs
AurevoirXavier: It doesn't use lock. It uses reserve. But we already free all reserve. So, delete it directly.

runtime/darwinia/src/lib.rs
AurevoirXavier: Only runtime call cares about the index. These won't affect the storage. I think there isn't security issue. Unless there are some on-going scheduled call. But those pallets were removed long time ago.

AurevoirXavier:
> I assume there will be another PR to handle the migration.

Yep.

pallet/account-migration/src/lib.rs
boundless-forest: Okay.

runtime/darwinia/src/lib.rs
boundless-forest: @hackfisher We have some dissicussion internally, what do you think?

hackfisher: Both options look good to me. I will probably keep them to ensure we maintain good co-coding conventions, as I will recall why we skipped these indexes in future. Additionally, I encourage others not to skip them by commenting as we have done previously, new contributors may be guided to bad conventions we didn't expect.

However, I do not have reasons to argue against it from a security perspective.

PR #1536, opened 2024-07-12
Add tracing support for the EthereumXcm pallet

### LINKED ISSUE #1535: Support `EthereumXcm` tracing
See https://github.com/moonbeam-foundation/moonbeam/pull/1801, we missed this part in our repo.

boundless-forest:
Fix #1535 

This PR is complete. However, I am wondering if it is still necessary to retain the EthereumXCM pallet. The pallet currently exists in the Darwinia and Koi networks, but the Crab network does not have it. I believe Msgport could be a suitable replacement in this scenario. If the pallet is not needed, this PR can be closed, and another one can be raised to delete the EthereumXCM. @hackfisher @AurevoirXavier @jiguantong


jiguantong (approved):

jiguantong:
EthereumXCM is used to call Darwinia Contract via XCM from other parachains. Previously, it was used to call Ethereum bridge from other parachains. I think we don’t need it now.

boundless-forest:
Okay. Let's remove it.

PR #1537, opened 2024-07-15
Deactivate async for Crab and Darwinia

AurevoirXavier:
EVM runtime test might fail.


runtime/crab/src/pallets/timestamp.rs
```diff
@@ -20,7 +20,7 @@
 use crate::*;
 
 impl pallet_timestamp::Config for Runtime {
-	type MinimumPeriod = ();
+	type MinimumPeriod = ConstU64<{ SLOT_DURATION / 2 }>;
```
boundless-forest, line 23: image
According to the documentation, the `MinimumPeriod` needs to be adjusted in Phase 3, along with the `AllowMultipleBlocksPerSlot`. Therefore, I suggest reverting this modification for the Crab and Darwinia runtimes.

AurevoirXavier: It's the revert.

boundless-forest: Okay. https://github.com/darwinia-network/darwinia/blob/v6.6.3/runtime/crab/src/pallets/timestamp.rs I checked the original config.

boundless-forest (approved):

PR #1545, opened 2024-07-22
impl get

boundless-forest:


pallet/ethtx-forwarder/src/lib.rs
```diff
@@ -259,8 +259,35 @@ impl<T: Config> Pallet<T> {
...
+			None,
+			None,
+			<T as pallet_evm::Config>::config(),
+		)
+		.map_err(|err| err.error.into())
+	}
+
 	#[cfg(feature = "evm-tracing")]
-	fn trace_tx(
+	fn trace_apply(
```
AurevoirXavier, line 290: Sounds strange; "trace on" or "trace at" sounds better.

boundless-forest: The primary rationale behind this is `T::ValidatedTransaction::apply`.  I aim to indicate that this is a tracing version of the apply function.

boundless-forest: Make sense. I keep the original naming.

PR #1548, opened 2024-07-25
Mutable contract address

AurevoirXavier:


pallet/staking/src/lib.rs
```diff
@@ -295,6 +291,17 @@ pub mod pallet {
...
 
+	/// KTON reward distribution contract address.
+	#[pallet::storage]
+	#[pallet::getter(fn kton_reward_distribution_contract)]
+	pub type KtonRewardDistributionContract<T: Config> =
+		StorageValue<_, T::AccountId, ValueQuery, KtonRewardDistributionContractDefault<T>>;
+	/// Default value for [`KtonRewardDistributionContract`].
+	#[pallet::type_value]
+	pub fn KtonRewardDistributionContractDefault<T: Config>() -> T::AccountId {
+		account_id()
```
boundless-forest, line 302: Why not set the actual distribution contract address `0x0DBFbb1Ab6e42F89661B4f98d5d0acdBE21d1ffC` as the default value? This way, we won't need to set the value through governance. It can also help avoid potential distribution reward errors during the storage set gap.

AurevoirXavier: Need to set this in the runtime upgrade.

This isn't a fixed address.

Due to some bugs in the contract side, Koi's contract needs to be upgraded, which is already different from other networks.

boundless-forest: Okay. I understand it now. Would you like to add a TODO or create an issue for that to avoid forgetting it?

PR #1554, opened 2024-08-08
Enable async backing

AurevoirXavier:


boundless-forest (approved):

runtime/crab/src/pallets/timestamp.rs
```diff
@@ -20,7 +20,7 @@
 use crate::*;
 
 impl pallet_timestamp::Config for Runtime {
-	type MinimumPeriod = ConstU64<{ SLOT_DURATION / 2 }>;
+	type MinimumPeriod = ();
```
boundless-forest, line 23: What's the difference between these two values?

AurevoirXavier: Async requires this to be set to 0.

PR #1578, opened 2024-08-26
Release `koi-6655`

AurevoirXavier:


pallet/staking/src/lib.rs
```diff
@@ -1202,14 +1208,18 @@ where
 		.map_err(|e| log::error!("failed to forward call due to {e:?}"))
 		.ok()
 		.and_then(|i| {
+			log::info!("getTopCollators({n})'s execution info {i:?}");
```
boundless-forest, line 1211: The log level should be debug or remove it.

AurevoirXavier: I don't think so.

This log is important and only available for non onchain-release build.

PR #1607, opened 2024-10-09
Clean issues

### LINKED ISSUE #1560: Disable lock/stake calls.

### LINKED ISSUE #1604: KtonStaking is not distributing rewards after latest runtime upgrade on Crab
https://crab-scan.darwinia.network/address/0x652182C6aBc0bBE41b5702b05a26d109A405EAcA?tab=txs

AurevoirXavier:
```
subalfred storage-key --pallet BridgeKusamaGrandpa --item ImportedHeaders
0xad61a448d2290042116be82bf61332ebf7e5937ed7be0ea26416b155b0109b48
```

- Remove Kusama header migration code
- Typo fixes
- Closes #1604
- Closes #1560
- Closes https://github.com/darwinia-network/KtonDAO/issues/23

image

### DIFF
--- a/runtime/crab/src/migration.rs
+++ b/runtime/crab/src/migration.rs
@@ -45,5 +45,9 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 }
 
 fn migrate() -> frame_support::weights::Weight {
-	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(0, 0)
+	if let Ok(k) = array_bytes::hex2bytes("0x1da53b775b270400e7e61ed5cbc5a146ab1160471b1418779239ba8e2b847e42d53de13b56da115d3342f0588bc3614108837de0ae21c270383d9f2de4db03c7b1314632314d8c74970d627c9b4f4c42e06688a9f7a2866905a810c4b1a49b8cb0dca3f1bc953905609869b6e9d4fb794cd36c5f") {
+		System::kill_storage(RuntimeOrigin::root(), vec![k]);
+	}
+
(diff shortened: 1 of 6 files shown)


runtime/crab/src/migration.rs
```diff
@@ -45,47 +45,5 @@ impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
 }
 
 fn migrate() -> frame_support::weights::Weight {
-	// dawinia
-	use darwinia_staking::CacheState;
-	if let Some(s) = migration::get_storage_value::<(CacheState, CacheState, CacheState)>(
-		b"DarwinaStaking",
```
hackfisher, line 51: It seems that here is also a typo, does it mean this migration was not successfully executed?

What's the impact for Crab.

hackfisher:
Please also complete step 2 in this issue to close it: https://github.com/darwinia-network/KtonDAO/issues/23

runtime/crab/src/migration.rs
AurevoirXavier: Get `None`, then skip the `put`.

Fallback to use the https://github.com/darwinia-network/darwinia/blob/1ac5283f5baaf351d472de87bc162d373ee4897e/pallet/staking/src/lib.rs#L244. This only affects the first session. The election might also be skipped.

hackfisher: If `ExposureCacheStates ` is renamed and not used anymore, should we clean its storage?

AurevoirXavier: There is no need for now. There are many unnecessary items. Clean them all up in Staking V3 at once.

PR #1619, opened 2024-11-04
Add new bootnodes

fewensa:
Add new backup bootnodes

- Darwinia
- Crab


AurevoirXavier:
Need to update the JSON file and revise the fetch logic. This change won't impact the cached JSON.

PR #1636, opened 2024-11-29
Improve deposits migration

### LINKED ISSUE #1635: Filter non-zero deposit items during migration

AurevoirXavier:
Closes #1635.

### DIFF
--- a/pallet/deposit/src/lib.rs
+++ b/pallet/deposit/src/lib.rs
@@ -40,7 +40,6 @@ pub use weights::WeightInfo;
 // core
 use core::marker::PhantomData;
 // crates.io
-use codec::FullCodec;
 use ethabi::{Function, Param, ParamType, StateMutability, Token};
 // darwinia
 use dc_types::{Balance, Moment};
@@ -50,11 +49,9 @@ use fp_evm::{CallOrCreateInfo, ExitReason};
 use frame_support::{
 	pallet_prelude::*,
 	traits::{Currency, ExistenceRequirement::AllowDeath, UnixTime},
-	PalletId,
 };
 use frame_system::pallet_prelude::*;
 use sp_core::H160;
-use sp_runtime::traits::AccountIdConversion;
 use sp_std::prelude::*;
 
 #[frame_support::pallet]
@@ -204,7 +201,7 @@ pub mod pallet {
 			let mut to_migrate = (0, Vec::new(), Vec::new());
 
 			// Take 0~10 deposits to migrate.
-			for d in deposits.by_ref().take(10) {
+			for d in deposits.by_ref().take(10).filter(|d| d.value != 0) {
 				if d.expired_time <= now {
 					to_claim.0 += d.value;
 					to_claim.1.push(d.id);
@@ -216,14 +213,13 @@ pub mod pallet {
 			}
 
 			if to_claim.0 != 0 {
-				T::Ring::transfer(&account_id(), who, to_claim.0, AllowDeath)?;
+				T::Ring::transfer(&T::Treasury::get(), who, to_claim.0, AllowDeath)?;
 				Self::deposit_event(Event::DepositsClaimed {
 					owner: who.clone(),
 					deposits: to_claim.1,
 				});
 			}
 			if to_migrate.0 != 0 {
-				T::Ring::transfer(&account_id(), &T::Treasury::get(), to_migrate.0, AllowDeath)?;
 				T::DepositMigrator::migrate(who.clone(), to_migrate.0, to_migrate.2)?;
 				Self::deposit_event(Event::DepositsMigrated {
 					owner: who.clone(),
@@ -337,12 +333,4 @@ where
 			_ => Err(<Error<T>>::MigrationFailedOnContract)?,
 		}
 	}
-}
-
-/// The account of the deposit pot.
-pub fn account_id<A>() -> A
-where
-	A: FullCodec,
-{
-	PalletId(*b"dar/depo").into_account_truncating()
 }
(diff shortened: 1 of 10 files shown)


runtime/darwinia/src/migration.rs
```diff
@@ -86,5 +87,11 @@ fn migrate() -> frame_support::weights::Weight {
 		]),
 	);
 
-	<Runtime as frame_system::Config>::DbWeight::get().reads_writes(r, w + 5)
+	let _ = Balances::transfer_all(
+		RuntimeOrigin::signed(PalletId(*b"dar/depo").into_account_truncating()),
```
hackfisher, line 91: The is still some CRAB left in deposit pallet account on Crab chain.

https://crab-scan.darwinia.network/address/0x6d6F646C6461722F6465706F0000000000000000

We also need to do the transfer migration for CRAB to clean it.

pallet/deposit/src/lib.rs
```diff
@@ -204,7 +201,7 @@ pub mod pallet {
 			let mut to_migrate = (0, Vec::new(), Vec::new());
 
 			// Take 0~10 deposits to migrate.
-			for d in deposits.by_ref().take(10) {
+			for d in deposits.by_ref().take(10).filter(|d| d.value != 0) {
```
hackfisher, line 204: Will this filter make the deposit with value=0 being not take out and still leaving in the deposits?

This could still cause `<Deposits<T>>::iter().drain().next()` still be stuck at the user with the deposit which value is zero?

https://github.com/darwinia-network/darwinia/blob/7ed9ae01f10a7d0cd99af9b6ac73de0518856bf8/pallet/deposit/src/lib.rs#L140-L141

AurevoirXavier: https://github.com/darwinia-network/darwinia/pull/1636/files#diff-90778f3c64239af47f21d89b2cbcd4d68857be8120e317604dfc69456180538dR111-R125

hackfisher:
@AurevoirXavier CI failed

pallet/deposit/src/lib.rs
```diff
@@ -216,14 +213,13 @@ pub mod pallet {
...
-				T::Ring::transfer(&account_id(), who, to_claim.0, AllowDeath)?;
+				T::Ring::transfer(&T::Treasury::get(), who, to_claim.0, AllowDeath)?;
 				Self::deposit_event(Event::DepositsClaimed {
 					owner: who.clone(),
 					deposits: to_claim.1,
 				});
 			}
 			if to_migrate.0 != 0 {
-				T::Ring::transfer(&account_id(), &T::Treasury::get(), to_migrate.0, AllowDeath)?;
 				T::DepositMigrator::migrate(who.clone(), to_migrate.0, to_migrate.2)?;
```
hackfisher, line 223: I'm thing if we fail with error here, it still end up with side effects like in line 216 (some of the funds might be transferred to user as we are in a loop)

We should emit a DepositMigration Failed events here?

AurevoirXavier: There will be an error log on node side.

Not sure if a Substrate event works for us nowadays. Only Darwinia has a Substrate explorer.

hackfisher: Let's emit Event anyway, at least we can trace back and index to it.

hackfisher: Fortunately, I remember that the value = 0 entry is this user’s first deposit; otherwise, we might have already encountered side effects.

pallet/deposit/src/lib.rs
```diff
@@ -216,14 +213,13 @@ pub mod pallet {
 			}
 
 			if to_claim.0 != 0 {
-				T::Ring::transfer(&account_id(), who, to_claim.0, AllowDeath)?;
+				T::Ring::transfer(&T::Treasury::get(), who, to_claim.0, AllowDeath)?;
```
hackfisher, line 218: Also emit failure event here if error.

pallet/deposit/src/lib.rs
```diff
@@ -216,14 +217,19 @@ pub mod pallet {
 			}
 
 			if to_claim.0 != 0 {
-				T::Ring::transfer(&account_id(), who, to_claim.0, AllowDeath)?;
+				if let Err(e) = T::Ring::transfer(&T::Treasury::get(), who, to_claim.0, AllowDeath)
+				{
+					Self::deposit_event(Event::MigrationFailedOnRefund);
+
+					Err(e)?;
```
hackfisher, line 224: Do not throw errors, otherwise there are still side effects l.

pallet/deposit/src/lib.rs
```diff
@@ -216,14 +217,19 @@ pub mod pallet {
...
+				}
+
 				Self::deposit_event(Event::DepositsClaimed {
 					owner: who.clone(),
 					deposits: to_claim.1,
 				});
 			}
 			if to_migrate.0 != 0 {
-				T::Ring::transfer(&account_id(), &T::Treasury::get(), to_migrate.0, AllowDeath)?;
 				T::DepositMigrator::migrate(who.clone(), to_migrate.0, to_migrate.2)?;
```
hackfisher, line 233: Should not throw errors here right?

AurevoirXavier: In FP, we treat any I/O operation as the side effect.

And here, especially we interact with the DB, `take` and `transfer` have already introduced the side effects. We should keep the error and utilize it to help rollback the state.

hackfisher: You didn't get what I mean.

If keep the error, this batch (10 items) will be inserted again, but some of them might have been migrated, then they will be keep migrated again and again.

hackfisher: Considering that you are on vacation and the likelihood of errors is very low, I suggest we keep things simple. At the same time, we must absolutely avoid any ledger inaccuracies or double-spending in case an error does occur.

> Then we can remove the event and insert the failure items into a different storage, and process that storage manually later. Using storage is easier to track and process compare to event.

If this looks simpler for you, It works for me. (Just need me take another several minutes to review the difference between them, good so far.)

hackfisher (approved):
Please bump version and release a new runtime for upgrade.

hujw77 (approved):
LGTM

PR #1652, opened 2025-01-15
Migration for weight trader

aurexav:


jiguantong (changes requested):
Koi also requires the same migration.

aurexav:
> Koi also requires the same migration.

Unless you can do some tests on Koi. 

Otherwise, it's meaningless. Koi runtime will be removed soon.

